PR #94, opened 2021-12-08
Enable test_shm_death_style

igchor:
It was missing from CMakeLists.txt


sathyaphoenix:
@igchor thanks for the PR. Looks like the update to OSS build was missed when the test was ported to a separate binary from the main suite. Thanks for fixing it. Once you sign the CLA, we'll be able to merge this.

PR #95, opened 2021-12-09
Fix cachebench for config with persistence enabled and NVM cache disabled

victoria-mcgrath:
Converted "NvmCacheState CacheAllocator::nvmCacheState_" to "std::optional<NvmCacheState>CacheAllocator::nvmCacheState_" to simplify NVM cache state handling when NVM cache is not enabled. This fixes an error message generated during cachebench run with cache persistence enabled and NVM cache disabled: "[...NvmCacheState.cpp:135] unable to deserialize nvm metadata file: no content in file: <path>/NvmCacheState".

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -3101,7 +3101,13 @@ std::optional<bool> CacheAllocator<CacheTrait>::saveNvmCache() {
     return false;
   }
 
-  nvmCacheState_.value().markSafeShutDown();
+  if (nvmCacheState_.has_value()) {
+    nvmCacheState_.value().markSafeShutDown();
+  }
+  else {
+    XLOG(ERR, "Could not mark NvmCacheState as safe shutdown.");
+    return false;
+  }
   return true;
 }
 
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -45,9 +45,7 @@ CacheAllocator<CacheTrait>::CacheAllocator(Config config)
           [this](Item* it) -> ItemHandle { return acquire(it); })),
       chainedItemLocks_(config_.chainedItemsLockPower,
                         std::make_shared<MurmurHash2>()),
-      cacheCreationTime_{util::getCurrentTimeSec()},
-      nvmCacheState_{config_.cacheDir, config_.isNvmCacheEncryptionEnabled(),
-                     config_.isNvmCacheTruncateAllocSizeEnabled()} {
+      cacheCreationTime_{util::getCurrentTimeSec()} {
   initCommon(false);
 }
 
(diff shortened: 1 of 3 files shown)


cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -3104,7 +3101,7 @@ std::optional<bool> CacheAllocator<CacheTrait>::saveNvmCache() {
     return false;
   }
 
-  nvmCacheState_.markSafeShutDown();
```
haowu14, line 3107: Please add an `if` block to check `nvmCacheState_` has value. Otherwise do an XLOG(ERR, "Could not mark NvmCacheState as safe shutdown.") and return false;
This shouldn't happen but it's good to have for safety.

cachelib/allocator/CacheAllocator.h
```diff
@@ -944,8 +944,17 @@ class CacheAllocator : public CacheBase {
   //
   // @return  time when the cache was created.
   time_t getCacheCreationTime() const noexcept { return cacheCreationTime_; }
+
+  // unix timestamp when the NVM cache was created. If NVM cahce isn't enaled,
+  // the cache creation time is returned instead.
```
haowu14, line 949: Thank you for adding the comment.

Please return 0 here if the nvm cache is not initialized just like before, and modify this comment accordingly. We don't want to change the existing behavior (though arguably your proposed behavior makes more sense) so that we can avoid migrations.

haowu14 (changes requested):
Hi!
Thank you for putting together this change. I left two minor comments and the rest looks great.
I'll import and merge it once the comments are addressed.

cachelib/allocator/CacheAllocator-inl.h
victoria-mcgrath: Done

cachelib/allocator/CacheAllocator.h
victoria-mcgrath: Done

cachelib/allocator/CacheAllocator-inl.h
victoria-mcgrath: Done

cachelib/allocator/CacheAllocator.h
victoria-mcgrath: Done

cachelib/allocator/CacheAllocator-inl.h
victoria-mcgrath: @haowu14 I made the changes that you requested but I noticed that CI workflow didn't pass on my fork. I'll work on this a little more and post an update here.

victoria-mcgrath: @haowu14 It turns out that the build is failing without my changes too.

haowu14: The failure that shows up in the build hook is the CLA. Please follow the instruction here https://github.com/facebook/CacheLib/pull/95#issuecomment-990067251 and after that I can import this for internal review. Thanks!

haowu14:
Hi @victoria-mcgrath . Sorry for the late reply and I hope you had a good break over the holidays.

During our internal tests we realized that this change still changes some behavior after all.
When constructing NvmCacheState object, we keep a filestream open at <cacheDir>/NvmCacheState. In this change, we do not create NvmCacheState object if NVM cache is not enabled. This breaks some of the assumptions elsewhere in CacheLib (mostly persistence) where we assume such a file to exist regardless of whether NVM cache is enabled.

Regarding to this error message of
```
unable to deserialize nvm metadata file: no content in file: /NvmCacheState
```

We will seek other way to avoid that and turn it into a warning.

PR #104, opened 2021-12-23
Disable pool rebalancing when cache eviction is disabled.

igchor:
testAllocateWithoutEviction was failing randomly at `XDCHECK(!config_.disableEviction);` in evictForSlabRelease.

I used the same approach as in different tests but I was also wondering if there should be some extra check (perhaps in config::validate?) to make sure poolRebalancer is not running when eviction is disabled.

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -379,7 +379,7 @@ CacheAllocator<CacheTrait>::allocateChainedItemInternal(
   (*stats_.allocAttempts)[pid][cid].inc();
 
   void* memory = allocator_->allocate(pid, requiredSize);
-  if (memory == nullptr) {
+  if (memory == nullptr && !config_.disableEviction) {
     memory = findEviction(pid, cid);
   }
   if (memory == nullptr) {
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -2351,7 +2351,7 @@ void CacheAllocator<CacheTrait>::releaseSlabImpl(
     const bool isMoved = moveForSlabRelease(releaseContext, item, throttler);
 
     // if moving fails, evict it
-    if (!isMoved) {
+    if (!isMoved && !config_.disableEviction) {
       evictForSlabRelease(releaseContext, item, throttler);
     }
     XDCHECK(allocator_->isAllocFreed(releaseContext, alloc));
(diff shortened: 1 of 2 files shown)


sathyaphoenix:
@igchor  Thanks for identifying this issue. PoolRebalancer also handles moving slabs when allocation failures are happening. The option to disableEviction does conflict with this intent. However, in such scenarios, I wonder if we can resort to only slab releases that can be done through moveForSlabRelease and not evictForSlabRelease instead. That is, if  moving can allocate and execute the slab release through free allocs in the MemoryAllocator. What do you think ? 

It sounds like you are bumping into this because of the corner cases with config.disableEviction. IMO, that option should be deprecated in CacheLib.  Do you also mind sharing some context on how you were lead to this issue. 

Also, I noticed the clang-format needs to be addressed before the PR can be pulled in.

igchor:
@sathyaphoenix  Yes, I think doing just moveForSlabRelease makes sense. I belive that disabling eviction and not providing moveCb will result in slabRelease always failing - should this be documented somewhere?

I encountered this issue after we did some changes inside allocate() related to https://github.com/facebook/CacheLib/discussions/102. Those changes probably resulted in different timings (the test failed randomly).

By the way, regarding the link above - I was also wondering what should be the behavior of disableEviction for a Cache with multiple memory tiers. Do you maybe have any thoughts on this? Should it only affect evictions from memory to SSD or also item movements between memory tiers (which is some ways similar to eviction - e.g. we also consult MMContainer)?

sathyaphoenix:
> Yes, I think doing just moveForSlabRelease makes sense. I believe that disabling eviction and not providing moveCb will result in slabRelease always failing - should this be documented somewhere?

That's a good point. My preference would be to deprecate the "disableEviction" option entirely since it poorly designed as is for hybrid cache and other cachelib features. Besides, how is it a cache if evictions are disabled :) Internally, we have one code reference to a system that might be using it. However, I believe they don't use it in production anymore. I am planning to deprecate this option all together.

igchor:
That sounds reasonable. Although I believe disableEviction might be quite useful for testing (especially for our team which is working on multi-tier implementation - this could allow us to just test the part of code responsible for moving items between memory tiers).

Anyhow, feel free to merge or close this PR. I've added a check for disableEviction before calling evictForSlabRelease as you suggested and also added a (missing I believe) check allocateChainedItemInternal.

sathyaphoenix:
@therealgymmy  does it make sense to deprecate this option as a part of the BE week  and close this PR out ?

therealgymmy:
@sathyaphoenix: we can't deprecate this yet as it's still used in one system in prod. They're migrating away. So for now we can rename it to "deprecated_disableEviction" and remove it after that use case is gone.

haowu14:
Just closing the loop here. CacheLib will be deprecating disableEviction (pending on internal migration).

PR #105, opened 2021-12-23
Do not reimplement getMaxAddressableSize in config.

igchor:
There is already a method within CompressedPtr that can
be used.


sathyaphoenix:
@agordon  the clang-format check failures seem to be related to installation of the clang-format in the github actions environment. can you please take a look.

Issue #125, opened 2022-03-22
Is it a bug of not checking against max bucket and lock power ?


wenhaocs:
There is a way of setting AccessConfig via number of estimated cache entries. However, unlike directly passing bucketPower and lockPower, the calculated bucketPower and lockPower is not checked against maximum. Is it a potential bug?
https://sourcegraph.com/github.com/facebook/CacheLib/-/blob/cachelib/allocator/CacheAllocatorConfig.h?L625

sathyaphoenix:
@wenhaocs  good catch. Do you mind sending a PR to address that and I can merge it.

wenhaocs:
@sathyaphoenix Please let me know if you want to throw error explicitly and tell users the limit for #cacheEntries, or just implicitly set the bucketPower to be maximum like in the PR.

jiayuebao:
@wenhaocs Shall we use std::min instead of std::max to pick the smaller value? And I noticed we didn't compare bucketSize with the maximum value as well, do we want to do such check? @sathyaphoenix

sathyaphoenix:
To be consistent with the constructor, it makes sense to throw an error when the value is going to be incorrect rather than silently fix it. Most likely in this case the user gave a wrong input and it is better to abort rather than silently run with a misconfigured hash table.  Also, like @jiayuebao  pointed out, we should be doing a std::min instead of max if we were to adjust to a meaningful value. 

@wenhaocs  can you add a simple unit test for this ? we can introduce a new .cpp [here](https://github.com/facebook/CacheLib/tree/affcd42574d0691f156e8bf5e9923316fb0d27f1/cachelib/allocator/tests) to add CacheAllocatorConfig unit tests. @jiayuebao  please let us know if there is a better place to add the test.

jiayuebao:
For unit test, we already have one [here](https://github.com/facebook/CacheLib/blob/affcd42574d0691f156e8bf5e9923316fb0d27f1/cachelib/allocator/tests/ChainedHashTest.cpp#L27). And we can add more cases there.

PR #130, opened 2022-03-30
Set limit for bucketPower on config with #cacheEntries

wenhaocs:
This PR fixes the potential bug of exceeding the limit for bucketPower and lockPower, by capping the bucketPower with kMaxBucketPower. 

Fix: https://github.com/facebook/CacheLib/issues/125


jiayuebao:
@wenhaocs Could you also remove `noexcept` keyword for the function?

wenhaocs:
> @wenhaocs Could you also remove `noexcept` keyword for the function?

Oh yeah, done! Generally, is there any check or process we can do locally, especially on mac?

PR #131, opened 2022-04-15
Extend cachbench with value validation

igchor:
The main purpose of this patch is to better simulate workloads in
cachebench. Setting validateValue to true allows to see performance
impact of using different mediums for memory cache.

### DIFF
--- a/cachelib/cachebench/cache/Cache-inl.h
+++ b/cachelib/cachebench/cache/Cache-inl.h
@@ -50,8 +50,10 @@ uint64_t Cache<Allocator>::fetchNandWrites() const {
 template <typename Allocator>
 Cache<Allocator>::Cache(const CacheConfig& config,
                         ChainedItemMovingSync movingSync,
-                        std::string cacheDir)
+                        std::string cacheDir,
+                        bool touchValue)
     : config_(config),
+      touchValue_(touchValue),
       nandBytesBegin_{fetchNandWrites()},
       itemRecords_(config_.enableItemDestructorCheck) {
   constexpr size_t MB = 1024ULL * 1024ULL;
@@ -397,6 +399,18 @@ typename Cache<Allocator>::WriteHandle Cache<Allocator>::insertOrReplace(
 }
 
 template <typename Allocator>
+void Cache<Allocator>::touchValue(const ReadHandle& it) const {
+  XDCHECK(touchValueEnabled());
+
+  auto ptr = reinterpret_cast<const uint8_t*>(getMemory(it));
+
+  /* The accumulate call is intended to access all bytes of the value
+   * and nothing more. */
+  auto sum = std::accumulate(ptr, ptr + getSize(it), 0ULL);
+  folly::doNotOptimizeAway(sum);
+}
+
+template <typename Allocator>
 typename Cache<Allocator>::ReadHandle Cache<Allocator>::find(Key key) {
   auto findFn = [&]() {
     util::LatencyTracker tracker;
@@ -406,6 +420,11 @@ typename Cache<Allocator>::ReadHandle Cache<Allocator>::find(Key key) {
     // find from cache and wait for the result to be ready.
     auto it = cache_->find(key);
     it.wait();
+
+    if (touchValueEnabled()) {
+      touchValue(it);
+    }
+
     return it;
   };
 
@@ -431,6 +450,11 @@ typename Cache<Allocator>::WriteHandle Cache<Allocator>::findToWrite(Key key) {
     // find from cache and wait for the result to be ready.
     auto it = cache_->findToWrite(key);
     it.wait();
+
+    if (touchValueEnabled()) {
+      touchValue(it);
+    }
+
     return it;
   };
 
--- a/cachelib/cachebench/cache/Cache.h
+++ b/cachelib/cachebench/cache/Cache.h
@@ -65,9 +65,11 @@ class Cache {
   //                      cache.
   // @param cacheDir      optional directory for the cache to enable
   //                      persistence across restarts.
+  // @param touchValue    read entire value on find
   explicit Cache(const CacheConfig& config,
                  ChainedItemMovingSync movingSync = {},
-                 std::string cacheDir = "");
+                 std::string cacheDir = "",
+                 bool touchValue = false);
 
   ~Cache();
 
@@ -179,6 +181,9 @@ class Cache {
     return getSize(item.get());
   }
 
+  // read entire value on find.
+  void touchValue(const ReadHandle& it) const;
+
   // returns the size of the item, taking into account ItemRecords could be
   // enabled.
   uint32_t getSize(const Item* item) const noexcept;
@@ -240,6 +245,9 @@ class Cache {
 
   // returns true if the consistency checking is enabled.
   bool consistencyCheckEnabled() const { return valueTracker_ != nullptr; }
+
+  // returns true if touching value is enabled.
+  bool touchValueEnabled() const { return touchValue_; }
 
   // return true if the key was previously detected to be inconsistent. This
   // is useful only when consistency checking is enabled by calling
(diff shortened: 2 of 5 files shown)


haowu14:
Do you mind sharing the use case scenario for this diff? If the intention is to validate that the value written is the same as value read, wouldn't consistency checking be enough?

igchor:
The main intention is to see how much overhead reading a value introduces when using different memory mediums (e g., DRAM vs. CXL-mem or PMEM). We would like to measure this overhead for different value sizes.

For check consistency, there is some additional overhead and, I think, also some additional requirements on values.

sathyaphoenix:
@haowu14  We discussed this in the call with Intel to merge PRs that are independent and useful. In this case, like @igchor  pointed out using the consistency checking mode to read the value adds other overheads in addition and the intent here is to trigger reading the item in pmem like scenarios where bandwidth is a concern for byte addressable medium.

cachelib/cachebench/cache/Cache.h
```diff
@@ -352,6 +361,9 @@ class Cache {
   // tracker for consistency monitoring.
   std::unique_ptr<ValueTracker> valueTracker_;
 
+  // expected value of all items in Cache.
+  std::optional<std::string> expectedValue_;
```
sathyaphoenix, line 365: This sounds like we are validating the value, but we are only validating a few bytes in each item. So either it is named incorrectly or we are intending to do something, but the code does not do all of it. 

What's the intent with adding this value validation ? My understanding was that this was to read the item's payload in entirety so that we can benchmark any DRAM or PMem bandwidth numbers. If that's the case, why are we reading only a subset of the item ?

cachelib/cachebench/cache/Cache.h
```diff
@@ -227,9 +230,15 @@ class Cache {
   // @param keys  list of keys that the stressor uses for the workload.
   void enableConsistencyCheck(const std::vector<std::string>& keys);
 
+  // enables validating all values on find. Each value is compared to
+  // expected Value.
+  void enableValueValidating(const std::string& expectedValue);
```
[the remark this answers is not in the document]
sathyaphoenix: In case of consistency checking, the enable happens later since the keys are generated by inserting into the cache and then we enable the consistency checking for the keys added to the cache. It is a bit convoluted, but we don't need to do the same.

sathyaphoenix (changes requested):

cachelib/cachebench/cache/Cache.h
igchor: Your understanding is correct. I believe we are verifying all bytes of the value right now. `expectedValue_` is being set to `hardcodedString_` which is 4MB. `The std::min<size_t>(expected.size(), getSize(it))` part is just to make sure that we are not reading outside of the item. This is analogous to: https://github.com/facebook/CacheLib/blob/main/cachelib/cachebench/cache/Cache-inl.h#L624

Is this reasoning correct?

cachelib/cachebench/cache/Cache.h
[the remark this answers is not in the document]
igchor: Inside Cache ctor we do not know what is the `expectedValue` yet. This is why I call `enableValueValidating` from CacheStressor which is responsible for generating the values (`hardcodedString_` for now).

cachelib/cachebench/cache/Cache.h
sathyaphoenix: > expectedValue_ is being set to hardcodedString_ which is 4MB.

The value is set by the caller (stressor) in a certain way, and we are leaking that into the cache. The fact that expectedValue[0 - item.getSize() -1] is the expected value is assumed in the cache. 

Taking a step back, we are not really building this feature to check item's content. That functionality is already handled by the consistency checking feature. We want this feature to simply read the item payload. Can this be done by the CacheStressor instead ? We also don't need to compare it against any expected value. As in, the result of that comparison is simply irrelevant. So there is no need to abort if the value is incorrect.

igchor: You are right that we are not really building this to check item's content. We came up with the idea of using memcmp to make sure that the code which reads the value is not optimized out by the compiler.

I don't think we can easily move the checking to the CacheStressor (all the latency checking code is done inside Cache). However, we can rewrite `validateValue` to not do any comparison. We can just iterate over the value and add a `volatile` somewhere to make sure the code is not optimized out (e.g. calculate the number of zero-bytes in the value and assign this number to a local volatile var). Would this be acceptable?

sathyaphoenix: @igchor  yeah, that seems more reasonable,  simpler and we don't need to think about terminating the run based on correctness. There is probably a way to do it without the volatile (check out folly::doNotOptimizeAway).

cachelib/cachebench/cache/Cache.h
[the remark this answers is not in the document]
igchor: I removed this function and added bool param to the Cache ctor - is it OK?

cachelib/cachebench/cache/Cache-inl.h
```diff
@@ -50,8 +50,10 @@ uint64_t Cache<Allocator>::fetchNandWrites() const {
 template <typename Allocator>
 Cache<Allocator>::Cache(const CacheConfig& config,
                         ChainedItemMovingSync movingSync,
-                        std::string cacheDir)
+                        std::string cacheDir,
+                        bool touchValue)
```
sathyaphoenix, line 54: can touchValue be part of the CacheConfig ?

cachelib/cachebench/cache/Cache-inl.h
```diff
@@ -397,6 +399,15 @@ typename Cache<Allocator>::ItemHandle Cache<Allocator>::insertOrReplace(
   return rv;
 }
 
+template <typename Allocator>
+void Cache<Allocator>::touchValue(const ItemHandle& it) const {
+  XDCHECK(touchValueEnabled());
+
+  auto ptr = reinterpret_cast<const uint8_t*>(getMemory(it));
+  auto sum = std::accumulate(ptr, ptr + getSize(it), 0ULL);
```
sathyaphoenix, line 407: leave a comment here that accumulate is intended to access all bytes of the item and nothing more.

cachelib/cachebench/cache/Cache-inl.h
```diff
@@ -408,13 +419,20 @@ typename Cache<Allocator>::ItemHandle Cache<Allocator>::find(Key key,
     // find from cache and wait for the result to be ready.
     auto it = cache_->findImpl(key, mode);
     it.wait();
+
```
sathyaphoenix, line 422: This code looks different from https://github.com/facebook/CacheLib/blob/69738cf1fe15b42f8db374aea5483496107ffcf2/cachelib/cachebench/cache/Cache-inl.h#L405 

Are you on a stale base commit for this PR ?

cachelib/cachebench/cache/Cache-inl.h
sathyaphoenix: nvm. I understand why it might not be a good idea to do that. We are really bending our back to add this code so that it fits under the place where find() api latency is measured and it is looking a bit convoluted. But that's fine for now.

sathyaphoenix (commented):
@igchor  the base rev seems incorrect. PTAL and update. Once that is done, I'll try to import it.

cachelib/cachebench/cache/Cache-inl.h
igchor: Done.

cachelib/cachebench/cache/Cache-inl.h
igchor: Done - rebased.

therealgymmy:
@igchor: can you rebase to latest trunk? We recently made some API change to CacheBench to use Read/Write handles. I will re-import again.

igchor:
@therealgymmy I've rebased, thanks

PR #132, opened 2022-04-15
Shorten critical section in findEviction

igchor:
Remove the item from mmContainer and drop the lock before attempting eviction.

The change improves throughput for default hit_ratio/graph_cache_leader_fbobj config by ~30%. It also reduces p99 latencies significantly. The improvement is even bigger for multi-tier approach (multiple memory tiers) which we are working on here: https://github.com/pmem/CacheLib

I was not able to find any races/synchronization problems with this approach but there is a good chance I missed something - it would be great if you could review and evaluate this patch.

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -1235,24 +1235,29 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
     Item* candidate = itr.get();
 
     // make sure no other thead is evicting the item
-    if (candidate->getRefCount() != 0) {
+    if (candidate->getRefCount() != 0 || candidate->isMoving()) {
       ++itr;
       continue;
     }
-  
-    incRef(*candidate);
+
+    if (candidate->isChainedItem()) {
+      candidate = &candidate->asChainedItem().getParentItem(compressor_);
+    }
+
+    auto toReleaseHandle = accessContainer_->find(*candidate);
+    if (!toReleaseHandle || !toReleaseHandle.isReady()) {
+      ++itr;
+      continue;
+    }
+
     itr.destroy();
 
     // for chained items, the ownership of the parent can change. We try to
     // evict what we think as parent and see if the eviction of parent
     // recycles the child we intend to.
-    auto toReleaseHandle = candidate->isChainedItem()
-                               ? tryEvictChainedItem(mmContainer, *candidate)
-                               : tryEvictRegularItem(mmContainer, *candidate);
-                               // XXX: fix chained item
-
-    if (toReleaseHandle) {
-      if (toReleaseHandle->hasChainedItem()) {
+    bool evicted = tryEvictItem(mmContainer, std::move(toReleaseHandle));
+    if (evicted) {
+      if (candidate->hasChainedItem()) {
         (*stats_.chainedItemEvictions)[pid][cid].inc();
       } else {
         (*stats_.regularItemEvictions)[pid][cid].inc();
@@ -1265,37 +1270,20 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
             toReleaseHandle->getConfiguredTTL().count());
       }
 
-      // we must be the last handle and for chained items, this will be
-      // the parent.
-      XDCHECK(toReleaseHandle.get() == candidate || candidate->isChainedItem());
-      XDCHECK_EQ(2u, toReleaseHandle->getRefCount());
-
-      // We manually release the item here because we don't want to
-      // invoke the Item Handle's destructor which will be decrementing
-      // an already zero refcount, which will throw exception
-      auto& itemToRelease = *toReleaseHandle.release();
-
       // Decrementing the refcount because we want to recycle the item
-      decRef(itemToRelease);
-      const auto ref = decRef(itemToRelease);
+      const auto ref = decRef(*candidate);
       XDCHECK_EQ(0u, ref);
 
       // check if by releasing the item we intend to, we actually
       // recycle the candidate.
       if (ReleaseRes::kRecycled ==
-          releaseBackToAllocator(itemToRelease, RemoveContext::kEviction,
+          releaseBackToAllocator(*candidate, RemoveContext::kEviction,
                                  /* isNascent */ false, candidate)) {
         return candidate;
       }
     }
 
-    // If we destroyed the itr to possibly evict and failed, we restart
-    // from the beginning again
-    if (!itr) {
-      itr.resetToBegin();
-      for (int i = 0; i < searchTries; i++)
-        ++itr;
-    }
+    itr.resetToBegin();
   }
   return nullptr;
 }
@@ -1349,9 +1337,10 @@ bool CacheAllocator<CacheTrait>::shouldWriteToNvmCacheExclusive(
 }
 
 template <typename CacheTrait>
-typename CacheAllocator<CacheTrait>::WriteHandle
-CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
-                                                Item& item) {
+bool CacheAllocator<CacheTrait>::tryEvictItem(MMContainer& mmContainer,
+                                              WriteHandle&& handle) {
+  auto& item = *handle;
+
   // we should flush this to nvmcache if it is not already present in nvmcache
   // and the item is not expired.
   const bool evictToNvmCache = shouldWriteToNvmCache(item);
@@ -1361,21 +1350,27 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
   // record the in-flight eviciton. If not, we move on to next item to avoid
   // stalling eviction.
   if (evictToNvmCache && !token.isValid()) {
-    stats_.evictFailConcurrentFill.inc();
-    decRef(item);
-    return WriteHandle{};
+    if (item.isChainedItem())
+      stats_.evictFailConcurrentFill.inc();
+    else
+      stats_.evictFailConcurrentFill.inc();
+    handle.reset();
+    return false;
   }
 
-  // If there are other accessors, we should abort. Acquire a handle here since
-  // if we remove the item from both access containers and mm containers
-  // below, we will need a handle to ensure proper cleanup in case we end up
-  // not evicting this item
+  // If there are other accessors, we should abort. We should be the only
+  // handle owner.
   auto evictHandle = accessContainer_->removeIf(item, &itemEvictionPredicate);
 
+  /* We got new handle so it's safe to get rid of the old one. */
+  handle.reset();
+
   if (!evictHandle) {
-    stats_.evictFailAC.inc();
-    decRef(item);
-    return evictHandle;
+    if (item.isChainedItem())
+      stats_.evictFailParentAC.inc();
+    else
+      stats_.evictFailAC.inc();
+    return false;
   }
 
   auto removed = mmContainer.remove(item);
@@ -1388,81 +1383,28 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
   // If the item is now marked as moving, that means its corresponding slab is
   // being released right now. So, we look for the next item that is eligible
   // for eviction. It is safe to destroy the handle here since the moving bit
-  // is set. Iterator was already advance by the remove call above.
+  // is set.
   if (evictHandle->isMoving()) {
-    stats_.evictFailMove.inc();
-    decRef(item);
-    return WriteHandle{};
+    if (item.isChainedItem())
+      stats_.evictFailParentMove.inc();
+    else
+      stats_.evictFailMove.inc();
+    return false;
   }
 
   // Ensure that there are no accessors after removing from the access
-  // container
-  XDCHECK(evictHandle->getRefCount() == 2);
+  // container.
+  XDCHECK(evictHandle->getRefCount() == 1);
 
   if (evictToNvmCache && shouldWriteToNvmCacheExclusive(item)) {
     XDCHECK(token.isValid());
     nvmCache_->put(evictHandle, std::move(token));
   }
-  return evictHandle;
-}
 
-template <typename CacheTrait>
-typename CacheAllocator<CacheTrait>::WriteHandle
-CacheAllocator<CacheTrait>::tryEvictChainedItem(MMContainer& mmContainer, Item& item) {
-  XDCHECK(item.isChainedItem());
+  /* Release the handle, item will be reused by the called. */
+  evictHandle.release();
 
-  ChainedItem* candidate = &item.asChainedItem();
-
-  // The parent could change at any point through transferChain. However, if
-  // that happens, we would realize that the releaseBackToAllocator return
-  // kNotRecycled and we would try another chained item, leading to transient
-  // failure.
-  auto& parent = candidate->getParentItem(compressor_);
-
-  const bool evictToNvmCache = shouldWriteToNvmCache(parent);
-
-  auto token = evictToNvmCache ? nvmCache_->createPutToken(parent.getKey())
-                               : typename NvmCacheT::PutToken{};
-
-  // if token is invalid, return. iterator is already advanced.
-  if (evictToNvmCache && !token.isValid()) {
-    stats_.evictFailConcurrentFill.inc();
-    return WriteHandle{};
-  }
-
-  // check if the parent exists in the hashtable and refcount is drained.
-  auto parentHandle =
-      accessContainer_->removeIf(parent, &itemEvictionPredicate);
-  if (!parentHandle) {
-    stats_.evictFailParentAC.inc();
-    return parentHandle;
-  }
-
-  // Ensure we have the correct parent and we're the only user of the
-  // parent, then free it from access container. Otherwise, we abort
-  auto removed = mmContainer.remove(item);
-  XDCHECK(removed);
-  XDCHECK_EQ(reinterpret_cast<uintptr_t>(&parent),
-             reinterpret_cast<uintptr_t>(parentHandle.get()));
-  XDCHECK_EQ(2u, parent.getRefCount()); // XXX?
-  XDCHECK(!parent.isInMMContainer());
-  XDCHECK(!parent.isAccessible());
-
-  // We need to make sure the parent is not marked as moving
-  // and we're the only holder of the parent item. Safe to destroy the handle
-  // here since moving bit is set.
-  if (parentHandle->isMoving()) {
-    stats_.evictFailParentMove.inc();
-    return WriteHandle{};
-  }
-
-  if (evictToNvmCache && shouldWriteToNvmCacheExclusive(*parentHandle)) {
-    XDCHECK(token.isValid());
-    XDCHECK(parentHandle->hasChainedItem());
-    nvmCache_->put(parentHandle, std::move(token));
-  }
-
-  return parentHandle;
+  return true;
 }
 
 template <typename CacheTrait>
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -1226,26 +1226,42 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
   // Keep searching for a candidate until we were able to evict it
   // or until the search limit has been exhausted
   unsigned int searchTries = 0;
-  auto itr = mmContainer.getEvictionIterator();
   while ((config_.evictionSearchTries == 0 ||
-          config_.evictionSearchTries > searchTries) &&
-         itr) {
+          config_.evictionSearchTries > searchTries)) {
     ++searchTries;
 
-    Item* toRecycle = itr.get();
+    Item* toRecycle = nullptr;
+    Item* candidate = nullptr;
 
-    Item* candidate =
-        toRecycle->isChainedItem()
-            ? &toRecycle->asChainedItem().getParentItem(compressor_)
-            : toRecycle;
+    mmContainer.withEvictionIterator(
+        [this, &candidate, &toRecycle, &searchTries](auto&& itr) {
+          while ((config_.evictionSearchTries == 0 ||
+                  config_.evictionSearchTries > searchTries) &&
+                 itr) {
+            ++searchTries;
 
-    // make sure no other thead is evicting the item
-    if (candidate->getRefCount() != 0 || !candidate->markMoving()) {
-      ++itr;
+            auto* toRecycle_ = itr.get();
+            auto* candidate_ =
+                toRecycle_->isChainedItem()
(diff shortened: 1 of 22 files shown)


igchor:
My main concern with this patch is that some thread might delete our item after we remove it from MMContainer and before calling tryEvictRegularItem. Perhaps we should increase the ref count?

therealgymmy:
@igchor: Hi, I will be reviewing this week. In the meanwhile, could you run the following cachebench tests? These test the consistency of item values that involve chained items.

My concern is around dropping the eviction lock for a regular early when a concurrent chained item eviction is happening, and vice-versa. An additional concern is around the interaction between Slab-release and eviction code path. I will focus on these during my review.

```
`cachelib/cachebench/test_configs/consistency/chained-items-stress-moving.json`
`cachelib/cachebench/test_configs/consistency/chained-items-stress-no-moving.json`
`cachelib/cachebench/test_configs/consistency/ram-with-deletes.json`
```

therealgymmy (commented):
@igchor:  Below are the scenarios I went through with this change. There is one potential bug (See (4) ).

When we drop the lock and try to remove an item from mm-container, what happens if:
1. another thread is trying to evict from the same mmcontainer?
	1. It can’t happen because only one thread can remove an item from AccessContainer. The concurrent thread will fail to remove this item from access-container due to refcount > 1. (Current thread is holding at least 1 refcount, and the concurrent thread must acquire a refcount before proceeding to remove from access-container.)
2. another thread is trying to evict one of its chained items from the same mmcontainer? And vice-versa?
	1. It can’t happen because only one thread can remove an item from AccessContainer. Either the current thread or the concurrent thread will skip this item.
3. another thread is trying to evict one of its chained items from a different mmcontainer? And vice-versa?
	1.  It can’t happen because only one thread can remove an item from AccessContainer. Either the current thread or the concurrent thread will skip this item.
4. [BUG] another thread is evicting a slab and trying to move/evict the same item? (CacheAllocator::evictForSlabRelease)
	1. Can an item be marked as moving concurrently as it was evicted from LRU?
		1. This is possible. The intent is that we cannot mark moving unless item is still in LRU. Dropping the lock changes this since we’re doing an atomic compare and exchange with acq_rel, but reading with relaxed.
			1. Please see “markMoving()” and “isMoving()” in CacheItem-inl.h and Refcount.h
			2. In slab release, we mark items as moving in markMovingForSlabRelease() in CacheAllocator-inl.h
		2. Fix is to make sure the atomic exchange and read are properly ordered. I’m forgetting details in the ordering, but first glance I think reading the moving bit with acquire and make sure we the failure scenario’s ordering in the exchange is also with acquire should do.
	2. Can an item be moved concurrently as it was evicted from LRU?
		1. Yes. See above.
	3. Can an item be evicted concurrently as it was evicted from LRU?
		1. Yes. See above.
5. another thread is evicting a slab and trying to move/evict one of its chained items? And vice versa?
	1. This cannot happen as the removal of the parent is synchronized on access container. Only one thread can remove the parent.

igchor:
@therealgymmy So, the only needed fix for correctness is to change the memory order on reads (at least for non-chained items)? Why the relaxed memory order is enough right now?

Also, here is my rationale for why 4th and 5th cases should work (perhaps after fixing memory order):

In my patch, if the item is removed from the LRU, then it must have also been removed from AccessContainer. This means that the item cannot be moved due to this check: https://github.com/facebook/CacheLib/blob/57c8bd34003fc71bbacef1fcce4e1d12fd5d09e6/cachelib/allocator/CacheAllocator-inl.h#L1075 . It can only be evicted.

If the item was already freed in findEviction (before any operation in SlabRelease), markMoving() will not be set (we see that the item is freed) and so, we can just finish.

There could be a problem when slabRlease calls markMoving() and then evictForSlabRelease just before: https://github.com/facebook/CacheLib/blob/57c8bd34003fc71bbacef1fcce4e1d12fd5d09e6/cachelib/allocator/CacheAllocator-inl.h#L1267. But at that time, the element is already removed from LRU, causing markMoving() to fail.

Also, if markMoving() is called after the element is removed from AccessContainer and before it is removed from LRU (inside tryEvictRegularItem) then the thread which is doing findEvition, will exit before freeing the item (because markMoving is set).

Are there any issues with my reasoning?

therealgymmy:
> the only needed fix for correctness is to change the memory order on reads (at least for non-chained items)? Why the relaxed memory order is enough right now?

@igchor: **please ignore my comments on the mem ordering.** I overlooked that after we remove from mm-container, we unmark an item's `kLinked` bit with acq_rel ordering, which will force the "eviction thread" and "slab release thread" to be correctly ordered. (Basically the existing code in L2397 `markMovingForSlabRelease`, and the code in L1370 `advanceIteratorAndTryEvictRegularItem`).

My concern previously was this:
1. slab release thread marks an item as moving as it sees item's kLinked bit is still marked
2. eviction thread sees the item moving bit UNmarked

I thought it could happen because L2397 will mark the bit with acq_rel but L1370 reads it with relaxed. However, the code in L1360 `mmContainer.remove(itr)` calls `Refcount::unmarkInMMContainer()` which unmarks the kLinked bit with acq_rel. This means L1360 and L2397 must be ordered correctly and thus L1370 which sequences-after L1360 must be ordered correctly with L2397 as well.

therealgymmy:
@igchor: can you rebase this PR to latest version? I will try importing it again and run through the internal tests.

therealgymmy:
@igchor: the following test failed:

```
cachelib/allocator/tests:cache-allocator-test - BaseAllocatorTest/3.AddChainedItemMultithread
```

It failed on this line: https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/tests/BaseAllocatorTest.h#L4522

With the following output:
```
buck-out/dev/gen/aab7ed39/cachelib/allocator/tests/cache-allocator-test#platform010-clang,private-headers/cachelib/allocator/tests/BaseAllocatorTest.h:4522
Expected: (nullptr) != (childItem), actual: (nullptr) vs nullptr
buck-out/dev/gen/aab7ed39/cachelib/allocator/tests/cache-allocator-test#platform010-clang,private-headers/cachelib/allocator/tests/BaseAllocatorTest.h:4522
Expected: (nullptr) != (childItem), actual: (nullptr) vs nullptr
buck-out/dev/gen/aab7ed39/cachelib/allocator/tests/cache-allocator-test#platform010-clang,private-headers/cachelib/allocator/tests/BaseAllocatorTest.h:4522
Expected: (nullptr) != (childItem), actual: (nullptr) vs nullptr
```

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1226,16 +1226,31 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+      ++itr;
+      continue;
+    }
+
+    if (candidate->isChainedItem()) {
+      candidate = &candidate->asChainedItem().getParentItem(compressor_);
+    }
+
+    auto toReleaseHandle = accessContainer_->find(*candidate);
+    if (!toReleaseHandle || !toReleaseHandle.isReady()) {
```
therealgymmy, line 1241: We don't need to check `isReady()` since this we still hold LRU lock and this item cannot have been evicted to NvmCache.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1332,37 +1330,44 @@ bool CacheAllocator<CacheTrait>::shouldWriteToNvmCacheExclusive(
...
   // record the in-flight eviciton. If not, we move on to next item to avoid
   // stalling eviction.
   if (evictToNvmCache && !token.isValid()) {
-    ++itr;
-    stats_.evictFailConcurrentFill.inc();
-    return WriteHandle{};
+    if (item.isChainedItem())
+      stats_.evictFailConcurrentFill.inc();
+    else
+      stats_.evictFailConcurrentFill.inc();
```
therealgymmy, line 1349: no need for `if`? we're bumping the same stat

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1226,16 +1226,31 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
     // evict what we think as parent and see if the eviction of parent
     // recycles the child we intend to.
-    auto toReleaseHandle =
-        itr->isChainedItem()
-            ? advanceIteratorAndTryEvictChainedItem(itr)
-            : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
-
-    if (toReleaseHandle) {
-      if (toReleaseHandle->hasChainedItem()) {
+    bool evicted = tryEvictItem(mmContainer, std::move(toReleaseHandle));
```
therealgymmy, line 1251: less error-prone if we keep the existing behavior of returning the `toReleaseHandle` and then later in this function we explicitly it release and decrement the refcount. (Near L1267)

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1332,37 +1330,44 @@ bool CacheAllocator<CacheTrait>::shouldWriteToNvmCacheExclusive(
...
   auto evictHandle = accessContainer_->removeIf(item, &itemEvictionPredicate);
 
+  /* We got new handle so it's safe to get rid of the old one. */
+  handle.reset();
+
   if (!evictHandle) {
-    ++itr;
-    stats_.evictFailAC.inc();
-    return evictHandle;
+    if (item.isChainedItem())
```
therealgymmy, line 1362: This cannot be a chained item since we're now always passing in the parent item handle into this function.

(We should find a way to keep this stat updated properly tho, maybe we can update this in `findEviction()` function instead)

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1371,101 +1376,28 @@ CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictRegularItem(
   // If the item is now marked as moving, that means its corresponding slab is
   // being released right now. So, we look for the next item that is eligible
   // for eviction. It is safe to destroy the handle here since the moving bit
-  // is set. Iterator was already advance by the remove call above.
+  // is set.
   if (evictHandle->isMoving()) {
-    stats_.evictFailMove.inc();
-    return WriteHandle{};
+    if (item.isChainedItem())
```
therealgymmy, line 1381: Same as L1362

therealgymmy (changes requested):
Made a pass at the latest version. I couldn't spot any obvious bugs in the code so far despite the chained item related test failure. Let me know if you can repro it on your own @igchor (I'll poke some more on my side in the meanwhile).

igchor:
@therealgymmy  Thank you for the feedback!

I managed to reproduce the issue locally and will work on fixing it.

> Did you mean under the hash-table lock? Seems like the change is to get an item handle by calling find()?

Yes, under the hash-table (Access Container) lock. I actually realized that instead of increasing the refCount (which requires this hash-table lock) we could just mark an item as moving in findEviction (if not already marked) This would prevent any other thread from releasing this item and we could even reuse the existing `evictNormalItemForSlabRelease` method (for ChainedItems we would still need separate code path). 

I have just one concern/question regarding this approach. In current implementation, is it possible for multiple threads to mark the same item as moving? (I don't see any check for that in markMoving()) If not, how is this prevented? Is there an assumption that only one thread can do SlabRelease at any point?

therealgymmy:
Hi Igor,

> I have just one concern/question regarding this approach. In current implementation, is it possible for multiple threads to mark the same item as moving?

Only one thread can mark the same item as moving. Multiple threads can be releasing slabs, but the same slab can only be released by one thread.

Thanks,
Jimmy
________________________________

igchor:
@therealgymmy I found out what the issue was: I was passing `candidate` to releaseBackToAllocator which after my changes represented the parent Item. We should pass pointer to the child we actually want to recycle instead.

Also, instead of increasing refCount of the item I now rely on `moving` flag. I changed the `markMoving` function to only succeed if the item is not yet marked as moving. This guarantees proper synchronization with other evicting threads and with SlabRelease thread. Once the item is marked as moving inside findEviction we can just execute the function which is used in SlabReleaseImpl for regular items - this removes some of the code duplication.

If the eviction fails (due to refCount being != 0) we unmark the item and check the flags. If flags are 0 this means that item is not used anywhere (and was not freed by any other thread since it was marked moving) and removed from AC and MMContainer by other thread so we can recycle it anyway.

After I implemented the above change I also realized it will be quite easy to use combined locking for eviction iterator which I also implemented (I took the idea from https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/MM2Q-inl.h#L256). The performance results are pretty good now, we can share the exact numbers with you once we finish running all benchmarks.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2727,6 +2627,11 @@ CacheAllocator<CacheTrait>::evictNormalItemForSlabRelease(Item& item) {
   auto token = evictToNvmCache ? nvmCache_->createPutToken(item.getKey())
                                : typename NvmCacheT::PutToken{};
 
+  if (skipIfTokenInvalid && evictToNvmCache && !token.isValid()) {
```
therealgymmy, line 2630: I think we can actually remove this logic. I don't remember clearly the rationale behind checking token validity right after creating it in `advanceIteratorAndTryEvictRegularItem()`.

Maybe it was to avoid evicting item pre-maturely (when we cannot insert it into nvm-cache). But this should be fairly rare in practice.

cachelib/allocator/Refcount.h
```diff
@@ -247,10 +247,10 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
   /**
    * The following four functions are used to track whether or not
    * an item is currently in the process of being moved. This happens during a
-   * slab rebalance or resize operation.
+   * slab rebalance or resize operation or during eviction.
    *
-   * An item can only be marked moving when `isInMMContainer` returns true.
-   * This operation is atomic.
+   * An item can only be marked moving when `isInMMContainer` returns true and
+   * the item is not yet marked as moving. This operation is atomic.
```
therealgymmy, line 253: hah, this is clever!

I will go through the code in more detail to reason about the logic change, but in theory this makes sense.

Minor comment. Please rename this API to `markExclusive`, since that is our intent now: marking an item as exclusive so cachelib can gain exclusive access to it internally.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1219,23 +1219,52 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
     auto toReleaseHandle =
-        itr->isChainedItem()
-            ? advanceIteratorAndTryEvictChainedItem(itr)
-            : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
+        evictNormalItem(*candidate, true /* skipIfTokenInvalid */);
+    auto ref = candidate->unmarkMoving();
 
-    if (toReleaseHandle) {
-      if (toReleaseHandle->hasChainedItem()) {
+    if (toReleaseHandle || ref == 0u) {
```
therealgymmy, line 1266: Why can we rely on just `ref == 0` to proceed with eviction?

Without a `toReleaseHandle`, it could mean we didn't remove the item (or its parent item) from access-container. It's not safe to evict it since it can still be accessed.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1269,15 +1302,18 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
       // recycle the candidate.
       if (ReleaseRes::kRecycled ==
           releaseBackToAllocator(itemToRelease, RemoveContext::kEviction,
-                                 /* isNascent */ false, candidate)) {
-        return candidate;
+                                 /* isNascent */ false, toRecycle)) {
+        return toRecycle;
+      }
+    } else if (ref == 0u) {
+      // it's safe to recycle the item here as there are no more
```
therealgymmy, line 1309: is this true? If `toReleaseHandle` is empty, we could have returned early from `evictNormalItem()` and didn't remove this item from access-container or mm-container.

cachelib/allocator/CacheAllocator-inl.h
igchor: Please see my answer below.

I actually think that we might destroy the `toReleaseHandle` first (unconditionally) and then only rely on `ref==0u`. `toReleaseHandle` will not actually release anything on destruction since we have the moving bit set. And if `unmarkMoving()` will return `0` we know that, either we or some other thread, managed to remove the item from AC and MMContainer and no one is using the item. I'll check this.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1219,23 +1219,52 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+
     // for chained items, the ownership of the parent can change. We try to
     // evict what we think as parent and see if the eviction of parent
     // recycles the child we intend to.
     auto toReleaseHandle =
-        itr->isChainedItem()
-            ? advanceIteratorAndTryEvictChainedItem(itr)
-            : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
+        evictNormalItem(*candidate, true /* skipIfTokenInvalid */);
+    auto ref = candidate->unmarkMoving();
```
therealgymmy, line 1264: Can you change the `const auto ref = decRef(itemToRelease);` at L1298 and change it to just `ref = ...;`?

Internal build is failing due to local shadow variable. (very inconvenient, but I cannot change the code of a github-backed PR.)

therealgymmy (changes requested):

cachelib/allocator/CacheAllocator-inl.h
therealgymmy: @igchor: ah I see. Yes you're right. And this is a good point too. In fact, after we have called unmarkMoving, if the ref is 0, then we must free/recycle the memory. Because whichever other thread that has removed it couldn't free the memory due to moving bit.

cachelib/allocator/Refcount.h
igchor: Hm, there is already a function `isExclusive` in Refcount.h Should we rename it to something else?

cachelib/allocator/CacheAllocator-inl.h
igchor: Done (in separate commit). Should I also remove the corresponding stat `evictFailConcurrentFill` or leave it for compatibility? 

Also this stat has the following comment next to it: "Eviction failures because this item has a potential concurrent fill from nvm cache. For consistency reason, we cannot evict it. Refer to NvmCache.h for more details."

cachelib/allocator/Refcount.h
therealgymmy: I think we can deprecate `isExclusive()` altogether. This change actually calls for a much bigger refactor in the slab release code path.

Moving bit is now denoting "exclusive" access to an item. Slab release should also be using it for gaining exclusive access, and the current notion of holding an active refcount to prevent someone else from removing/evicting an item can be fully deprecated.

Will elaborate in separate places about slab release.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2721,7 +2601,7 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
 
 template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::WriteHandle
-CacheAllocator<CacheTrait>::evictNormalItemForSlabRelease(Item& item) {
+CacheAllocator<CacheTrait>::evictNormalItem(Item& item) {
```
therealgymmy, line 2604: In L2620, do we still need `accessContainer->remoevIf` to return us a handle?

It seems we just need to return `true/false`.

cachelib/allocator/MM2Q-inl.h
```diff
@@ -250,22 +250,19 @@ MM2Q::Container<T, HookPtr>::getEvictionIterator() const noexcept {
...
+  // to get advantage of combining, use withEvictionIterator
   LockHolder l(*lruMutex_);
   return Iterator{std::move(l), lru_.rbegin()};
 }
 
+template <typename T, MM2Q::Hook<T> T::*HookPtr>
+template <typename F>
+void MM2Q::Container<T, HookPtr>::withEvictionIterator(F&& fun) {
+  lruMutex_->lock_combine([this, &fun]() {
+    fun(Iterator{LockHolder{}, lru_.rbegin()});
```
therealgymmy, line 262: Do we have other call sites of EvictionIterator that requires a lock being held?

In this design, we no longer exposes eviction iterator outside MM-container, so I think we can simplify it to no longer require a lock passed into its constructor.

----

I.e. can we delete `getEvictionIterator()`?

therealgymmy (commented):
Changes look great! I left high-level comments on simplifying eviction/slab-release further.

Internal unit tests have passed. I will move on to testing this on prod systems.

---

Do you have latest numbers of cachebench throughput improvement before/after this change?

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2664,7 +2544,7 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
     auto owningHandle =
         item.isChainedItem()
             ? evictChainedItemForSlabRelease(item.asChainedItem())
```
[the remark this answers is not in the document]
therealgymmy: > Would you be OK with doing this refactor for Slab Release in separate PR (after this one would be merged)?

Yes that is totally fine.

> One specific concern with the approach you suggested is: when should we call nvmCache->put  If we would create putToken before removing the item from AccessContainer and MMContainer and issued the actual put after item.isOnlyMoving() it will result in keeping they key in the inFlightPut map longer - is this OK?

Oh we should issue a put to NvmCache earlier. As soon as we have removed the RAM parent item from access-container. This is because from DRAM cahce's perspective, it is already invisible and it's expected that this item is on the queue to get inserted into NvmCache. (Similar to the existing logic in evictChainedItemForSlabRelease() )

If we delay the insertion to NvmCache until after "isOnlyMoving() == true", we actually become more at risk of seeing a concurrent Lookup to the same key. In which case, we will be forced to give up the insertion altogether. (Since the concurrent lookup have to return quickly to inform the caller a key exists or not, and in this case we don't have the item inserted in NvmCache yet and we will be forced to drop the insertion altogether). This is to avoid a scenario like:

```
(1) T1 -> Created PutToken (not yet enqueued or still in the queue)
(2) T2 -> GET -> Miss
(3) T1 -> Inserted into NvmCache
(4) T2 -> GEt -> Hit <=== Bad. This is read/write inconsistency. This should return a miss as well.
```

Today in our logic, we handle this correctly by returning a miss at (4), but it's not ideal. Ideally we want (1) and (3) to be as close as possible to avoid (2) getting in between. Increasing the time gap between (1) and (3) will make (2) more likely to get interleaved.

igchor:
@therealgymmy I have opened a new PR: https://github.com/facebook/CacheLib/pull/166 which basically a subset of this one. It would be great if you could take a look.

PR #136, opened 2022-04-29
Update Item_and_Handle.md

xiedeyantu:
WriteHandle -> ItemHandle ?


jiayuebao:
@xiedeyantu Nothing need to change here. ReadHandle and WriteHandle are introduced to take the replace of ItemHandle. ReadHandle is a read-only handle. WriteHandle is for both read & write, which is very similar to ItemHandle thus we can think of ItemHandle as the old name of a WriteHandle.

PR #138, opened 2022-05-02
First set of changes to cache configuration API to enable multi-tier caches

victoria-mcgrath:
These changes introduce per-tier cache configuration required to implement features discussed here: https://github.com/facebook/CacheLib/discussions/102. These specific changes enable single DRAM tier configs only which are compatible with the current version of cachelib. Configuration API will be expanded as multi-tier changes in other parts of the library are introduced.


cachelib/allocator/MemoryTierCacheConfig.h
```diff
@@ -0,0 +1,72 @@
...
+  // tier is a half of the total cache size, set both tiers' ratios to 1.
+  size_t ratio{0};
+
+ private:
+  // TODO: introduce a container for tier settings when adding support for
+  // file-mapped memory
+  MemoryTierCacheConfig() = default;
+};
+} // namespace cachelib
+} // namespace facebook
```
jiayuebao, line 72: Could you add a new line?

cachelib/allocator/MemoryTierCacheConfig.h
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  // Specifies ratio of this memory tier to other tiers. Absolute size
+  // of each tier can be calculated as:
+  // cacheSize * tierRatio / Sum of ratios for all tiers; the difference
+  // between total cache size and sum of all tier sizes resulted from
+  // round off error is accounted for when calculating the last tier's
+  // size to make the totals equal.
+  MemoryTierCacheConfig& setRatio(double _ratio) {
+    ratio = _ratio;
```
jiayuebao, line 48: A "double" is passed here while ratio is of "size_t" type.

jiayuebao (changes requested):
Could you fix the build failure by adding `MemoryTierCacheConfig.h` to "cachelib/allocator/TARGETS" and `CacheAllocatorConfigTest.cpp` to "cachelib/allocator/tests/TARGETS"?

cachelib/allocator/MemoryTierCacheConfig.h
victoria-mcgrath: Done

cachelib/allocator/MemoryTierCacheConfig.h
victoria-mcgrath: Done

victoria-mcgrath:
> Could you fix the build failure by adding `MemoryTierCacheConfig.h` to "cachelib/allocator/TARGETS" and `CacheAllocatorConfigTest.cpp` to "cachelib/allocator/tests/TARGETS"?

@jiayuebao I don't seem to have access to TARGETS files. Could you give me more specific instruction how I can address your request?

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -199,6 +201,25 @@ class CacheAllocatorConfig {
...
+  // Each vector element describes configuration for a single memory cache tier.
+  // Tiers can be set up as ratios of total cache size or have their sizes
+  // explicitly specified. If ratios are used, then total cache size must be set
+  // before this method is called; if sizes are explicitly specified, then
+  // their sum sizes must match the total cache size if it's previously set; if
+  // the total size has not been set then this method will set it as a sum of
+  // all tier sizes
+  CacheAllocatorConfig& configureMemoryTiers(const MemoryTierConfigs& configs);
+
+  // Sets total cache size and configures cache memory tiers. This method
```
sathyaphoenix, line 214: does this override the previously set cache size ? And by cache size we mean the DRAM cache size ?

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -199,6 +201,25 @@ class CacheAllocatorConfig {
   // cachePersistence()
   CacheAllocatorConfig& usePosixForShm();
 
+  // Configures cache memory tiers. Accepts vector of MemoryTierCacheConfig.
```
sathyaphoenix, line 204: Is there a description of what Memory tiers refers to here ? There is already the notion of tiers through hybrid cache. It would help if we can point to some documentation about the memory tier here.

jiayuebao:
@victoria-mcgrath I just realized the TARGETS files aren't synced to the Github. Let me check whether we shall modify the file internally on our end.

cachelib/allocator/CacheAllocatorConfig.h
victoria-mcgrath: I updated this method's description to list possible memory types for memory tiers. In the future, we could convert this write-up https://github.com/facebook/CacheLib/discussions/102 into a documentation section to refer to.

cachelib/allocator/CacheAllocatorConfig.h
victoria-mcgrath: By cache size we mean "total cache size" which is the sum of all memory tier sizes. By default there's only one DRAM memory tier whose size matches total cache size.

I decided to delete this 2 argument configureMemoryTiers method.

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -597,7 +630,8 @@ CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::setCacheName(
 }
 
 template <typename T>
-CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::setCacheSize(size_t _size) {
+CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::setCacheSizeImpl(
```
jiayuebao, line 633: Shall we not change `setCacheSize` function as other users are using this function without setting the multi-tier cache config?  Can we add a new setCacheSize function only for multi-tier caches?

jiayuebao (changes requested):

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -606,6 +640,18 @@ CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::setCacheSize(size_t _size) {
   return *this;
 }
 
+template <typename T>
+CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::setCacheSize(size_t _size) {
+  if (memoryTierConfigs.size() == 1) {
+    memoryTierConfigs[0].setSize(size);
```
jiayuebao, line 646:
```suggestion
    memoryTierConfigs[0].setSize(_size);
```
Is it a typo?

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -1001,6 +1136,9 @@ const CacheAllocatorConfig<T>& CacheAllocatorConfig<T>::validate() const {
     throw std::invalid_argument(
         "It's not allowed to enable both RemoveCB and ItemDestructor.");
   }
+
+  validateMemoryTierSizes();
```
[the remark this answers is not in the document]
victoria-mcgrath: @jiayuebao I fixed failing tests by updating default DRAM tier to have size that matches default total cache size. In general, changes in this PR shouldn't break current DRAM cache configuration. I ran a subset of unit tests locally and confirmed that they passed. Please let me know of remaining failing tests if any.

jiayuebao:
Hi @victoria-mcgrath: The change will still break tons of our internal tests. The reason is not all cachelib user cases are calling `setCacheSize()` to configure the cache size; some of them are using `config.size = xxx`. In this case, could you try not to directly change the existing cachelib APIs but introduce new ones for the change? 

BTW, can I know whether the test failure is visible from your side?

victoria-mcgrath:
> Hi @victoria-mcgrath: The change will still break tons of our internal tests. The reason is not all cachelib user cases are calling `setCacheSize()` to configure the cache size; some of them are using `config.size = xxx`. In this case, could you try not to directly change the existing cachelib APIs but introduce new ones for the change?
> 
> BTW, can I know whether the test failure is visible from your side?

@jiayuebao I did overlook a bunch of test changes. Thanks for pointing it out. I am working on fixing them.

jiayuebao:
Hey @victoria-mcgrath any update on the fix?

victoria-mcgrath:
> Hey @victoria-mcgrath any update on the fix?

I am still working on this.

victoria-mcgrath:
@jiayuebao I just discussed this problem with @igchor and he had a much better idea to work around this problem: we can only allow setting ratios of tiers and calculate their sizes only when we actually allocate memory. Then, the timing and method of setting the total cache size shouldn't seem to matter. In our internal experiments, we never needed to specify tier sizes explicitly, so this has been so far a somewhat redundant feature anyways. We'll work on the change and share it here.

victoria-mcgrath:
@jiayuebao and @igchor: please review the latest changes that disable size property for tiers.

jiayuebao:
> @jiayuebao and @igchor: please review the latest changes that disable size property for tiers.

Will let @igchor review first. If things look good, I will re-import the change again.

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -1028,6 +1080,23 @@ bool CacheAllocatorConfig<T>::validateStrategy(
          (type != PoolOptimizeStrategy::MarginalHits || trackTailHits);
 }
 
+template <typename T>
+bool CacheAllocatorConfig<T>::validateMemoryTiers() const {
```
igchor, line 1084: nit: shouldn't it return const CacheAllocatorConfig<T>& as normal validate?

igchor (commented):
Looks good to me

PR #141, opened 2022-05-25
Tracking per-pool usage stats

guptask:
This per pool size statistics reports the current used space in each memory pool. Pool Usage = PoolStats->poolUsableSize - PoolStats->freeMemoryBytes(). This statistics will be reported in the final report generated by cachebench.

### DIFF
--- a/cachelib/cachebench/cache/CacheStats.h
+++ b/cachelib/cachebench/cache/CacheStats.h
@@ -32,6 +32,8 @@ struct Stats {
   uint64_t allocAttempts{0};
   uint64_t allocFailures{0};
 
+  std::vector<double> poolUsageFraction;
+
   uint64_t numCacheGets{0};
   uint64_t numCacheGetMiss{0};
   uint64_t numRamDestructorCalls{0};
@@ -115,6 +117,12 @@ struct Stats {
                           invertPctFn(allocFailures, allocAttempts))
         << std::endl;
     out << folly::sformat("RAM Evictions : {:,}", numEvictions) << std::endl;
+
+    for (auto pid = 0U; pid < poolUsageFraction.size(); pid++) {
+      out << folly::sformat("Fraction of pool {:,} used : {:.2f}", pid,
+                            poolUsageFraction[pid])
+          << std::endl;
+    }
 
     if (numCacheGets > 0) {
       out << folly::sformat("Cache Gets    : {:,}", numCacheGets) << std::endl;
--- a/cachelib/cachebench/cache/Cache-inl.h
+++ b/cachelib/cachebench/cache/Cache-inl.h
@@ -494,15 +494,23 @@ bool Cache<Allocator>::checkGet(ValueTracker::Index opId,
 template <typename Allocator>
 Stats Cache<Allocator>::getStats() const {
   PoolStats aggregate = cache_->getPoolStats(pools_[0]);
+  auto usageFraction =
+      1.0 - (static_cast<double>(aggregate.freeMemoryBytes())) /
+                aggregate.poolUsableSize;
+  Stats ret;
+  ret.poolUsageFraction.push_back(usageFraction);
   for (size_t pid = 1; pid < pools_.size(); pid++) {
-    aggregate += cache_->getPoolStats(static_cast<PoolId>(pid));
+    auto poolStats = cache_->getPoolStats(static_cast<PoolId>(pid));
+    usageFraction = 1.0 - (static_cast<double>(poolStats.freeMemoryBytes())) /
+                              poolStats.poolUsableSize;
+    ret.poolUsageFraction.push_back(usageFraction);
+    aggregate += poolStats;
   }
 
   const auto cacheStats = cache_->getGlobalCacheStats();
   const auto rebalanceStats = cache_->getSlabReleaseStats();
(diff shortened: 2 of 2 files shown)


therealgymmy:
@guptask: Can you clarify the goal of this change? The change here only exports how many bytes is allocated towards each memory pool. (It doesn't track per allocation class usage).

For pool size, you should already be able to get it via `CacheAllocator::getPoolStats(poolId)`. And then look at `PoolStats::poolSize`. (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L154)

For per-allocation-class size, you can refer to `PoolStats::numSlabsForClass(classId)` (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L213).

guptask:
> @guptask: Can you clarify the goal of this change? The change here only exports how many bytes is allocated towards each memory pool. (It doesn't track per allocation class usage).
> 
> For pool size, you should already be able to get it via `CacheAllocator::getPoolStats(poolId)`. And then look at `PoolStats::poolSize`. (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L154)
> 
> For per-allocation-class size, you can refer to `PoolStats::numSlabsForClass(classId)` (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L213).

Are you referring to the `poolUsedSize` metric ? This metric tracks the capacity of a pool that is currently under use. The purpose of this change is to track how much capacity of each pool is getting utilized.
It is not the same as the `PoolStats` metrics `poolSize` or `poolUsableSize`.
I added the `getAllocationSize` API to calculate the pool used size as a check in the `AllocatorHitStats` unit test.

therealgymmy:
@guptask: `memoryPoolSize_[poolId]` gives you the size of bytes that have been allocated by the memory pool. It doesn't distinguish between what's allocated for items and what's in freed list inside the pool. Is this what you want? (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/memory/SlabAllocator.cpp#L362)

You can already get this stat by checking `CacheStats::mpStats::allocatedSlabs()`  (https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L168)

therealgymmy:
>  The poolUsedSize statistic is supposed to track the pool usage. Based on my understanding now, pool_used_size = Slab::ksize * (MPStats->freeSlabs + class_aggregate of ACStats->usedSlabs).
allocatedSlabs(), on the other hand, equals MPStats->freeSlabs + class_aggregate of ACStats->usedSlabs and ACStats->freeSlabs. Does this look correct to you ?

@guptask: Hmm I don't fully follow. Do you want to know a memory pool's amount of bytes that are currently allocated for items in cache? If so, you should exclude free memory that are nonetheless "allocated" for this pool. MPStats->freeSlabs and aggregate of ACStats->freeSlabs both are considered free memory. You also want to exclude ACStats->freeAllocs.

You can get all of this by looking at the `PoolStats` class. You can take the PoolStats->poolUsableSize and subtract from it PoolStats->freeMemoryBytes(). https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L190

The computation of freeMemoryBytes() is in MPStats class, and can be found here: https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/memory/MemoryAllocatorStats.h#L104

guptask:
> > The poolUsedSize statistic is supposed to track the pool usage. Based on my understanding now, pool_used_size = Slab::ksize * (MPStats->freeSlabs + class_aggregate of ACStats->usedSlabs).
> > allocatedSlabs(), on the other hand, equals MPStats->freeSlabs + class_aggregate of ACStats->usedSlabs and ACStats->freeSlabs. Does this look correct to you ?
> 
> @guptask: Hmm I don't fully follow. Do you want to know a memory pool's amount of bytes that are currently allocated for items in cache? If so, you should exclude free memory that are nonetheless "allocated" for this pool. MPStats->freeSlabs and aggregate of ACStats->freeSlabs both are considered free memory. You also want to exclude ACStats->freeAllocs.
> 
> You can get all of this by looking at the `PoolStats` class. You can take the PoolStats->poolUsableSize and subtract from it PoolStats->freeMemoryBytes(). https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheStats.h#L190
> 
> The computation of freeMemoryBytes() is in MPStats class, and can be found here: https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/memory/MemoryAllocatorStats.h#L104

Thanks. I'll update the PR accordingly.

cachelib/cachebench/cache/CacheStats.h
```diff
@@ -116,6 +118,12 @@ struct Stats {
         << std::endl;
     out << folly::sformat("RAM Evictions : {:,}", numEvictions) << std::endl;
 
+    for (auto pid = 0U; pid < poolUsageFraction.size(); pid++) {
```
therealgymmy, line 121: This is only tracking per-pool usage stats right? Can you update the PR title to reflect this?

guptask: done

PR #144, opened 2022-06-06
Fix simple_cache build error(because of handle mismatch).

ajaysjoshi:
Simple cache build fails with error:

```
CacheItemHandle get(CacheKey key) { return gCache_->find(key); }
      |                                            ~~~~~~~~~~~~~^~~~~
      |                                                         |
      |                                                         facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait>::ReadHandle {aka facebook::cachelib::detail::ReadHandleImpl<facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait> >}
```

Change the Itemhandle to use the correct member aka. ReadHandle.


therealgymmy:
Thanks for sending in the fix!

ajaysjoshi:
@therealgymmy Thanks for the review.
I think I have already signed the CLA, is there something else I need to do as well?

examples/simple_cache/main.cpp
```diff
@@ -22,7 +22,7 @@ namespace cachelib_examples {
 using Cache = cachelib::LruAllocator; // or Lru2QAllocator, or TinyLFUAllocator
 using CacheConfig = typename Cache::Config;
 using CacheKey = typename Cache::Key;
-using CacheItemHandle = typename Cache::ItemHandle;
+using CacheItemHandle = typename Cache::ReadHandle;
```
therealgymmy, line 25: Can you rename this to `CacheReadHandle`? Also need to update L47 where CacheItemHandle is used. (We have since split ItemHandle into Read/Write handle to enforce r/w semantics).

therealgymmy (changes requested):

examples/simple_cache/main.cpp
ajaysjoshi: Thanks, I have made the changes, can you reimport the pull request?

PR #166, opened 2022-10-11
Prepare findEviction to be extended for multiple memory tiers

igchor:
This PR refactors the eviction logic (inside `findEviction`) so that it will be easy to add support for multiple memory tiers.  Problem with multi-tier configuration is that the critical section under MMContainer lock is too long. To fix that we have implemented an algorithm which utilize WaitContext to decrease the critical section. (which will be part of future PRs).

The idea is to use `moving` (now `exclusive`) bit to synchronize eviction with SlabRelease (and in future, with readers). In this PR, I only changed how `findEviction` synchronizes with SlabRelease.

This PR is a subset of: https://github.com/facebook/CacheLib/pull/132 
The above PR introduced some performance regressions in the single-memory-tier version which we weren't able to fix yet, hence we decided to first implement this part (which should not affect performance) and later we can add separate path for multi-tier or try to optimize the original patch.

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -1228,15 +1228,23 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
     (*stats_.evictionAttempts)[pid][cid].inc();
 
     Item* candidate = itr.get();
-    mmContainer.remove(itr);
+
+    // make sure no other thead is evicting the item
+    if (candidate->getRefCount() != 0) {
+      ++itr;
+      continue;
+    }
+  
+    incRef(*candidate);
     itr.destroy();
 
     // for chained items, the ownership of the parent can change. We try to
     // evict what we think as parent and see if the eviction of parent
     // recycles the child we intend to.
     auto toReleaseHandle = candidate->isChainedItem()
-                               ? tryEvictChainedItem(*candidate)
+                               ? tryEvictChainedItem(mmContainer, *candidate)
                                : tryEvictRegularItem(mmContainer, *candidate);
+                               // XXX: fix chained item
 
     if (toReleaseHandle) {
       if (toReleaseHandle->hasChainedItem()) {
@@ -1255,7 +1263,7 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
       // we must be the last handle and for chained items, this will be
       // the parent.
       XDCHECK(toReleaseHandle.get() == candidate || candidate->isChainedItem());
-      XDCHECK_EQ(1u, toReleaseHandle->getRefCount());
+      XDCHECK_EQ(2u, toReleaseHandle->getRefCount());
 
       // We manually release the item here because we don't want to
       // invoke the Item Handle's destructor which will be decrementing
@@ -1263,6 +1271,7 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
       auto& itemToRelease = *toReleaseHandle.release();
 
       // Decrementing the refcount because we want to recycle the item
+      decRef(itemToRelease);
       const auto ref = decRef(itemToRelease);
       XDCHECK_EQ(0u, ref);
 
@@ -1275,9 +1284,13 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
       }
     }
 
-    // Insert item back to the mmContainer if eviction failed.
-    mmContainer.add(*candidate);
-    itr.resetToBegin();
+    // If we destroyed the itr to possibly evict and failed, we restart
+    // from the beginning again
+    if (!itr) {
+      itr.resetToBegin();
+      for (int i = 0; i < searchTries; i++)
+        ++itr;
+    }
   }
   return nullptr;
 }
@@ -1344,6 +1357,7 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
   // stalling eviction.
   if (evictToNvmCache && !token.isValid()) {
     stats_.evictFailConcurrentFill.inc();
+    decRef(item);
     return WriteHandle{};
   }
 
@@ -1355,9 +1369,12 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
 
   if (!evictHandle) {
     stats_.evictFailAC.inc();
+    decRef(item);
     return evictHandle;
   }
 
+  auto removed = mmContainer.remove(item);
+  XDCHECK(removed);
   XDCHECK_EQ(reinterpret_cast<uintptr_t>(evictHandle.get()),
              reinterpret_cast<uintptr_t>(&item));
   XDCHECK(!evictHandle->isInMMContainer());
@@ -1369,12 +1386,13 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
   // is set. Iterator was already advance by the remove call above.
   if (evictHandle->isMoving()) {
     stats_.evictFailMove.inc();
+    decRef(item);
     return WriteHandle{};
   }
 
   // Ensure that there are no accessors after removing from the access
   // container
-  XDCHECK(evictHandle->getRefCount() == 1);
+  XDCHECK(evictHandle->getRefCount() == 2);
 
   if (evictToNvmCache && shouldWriteToNvmCacheExclusive(item)) {
     XDCHECK(token.isValid());
@@ -1385,7 +1403,7 @@ CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
 
 template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::WriteHandle
-CacheAllocator<CacheTrait>::tryEvictChainedItem(Item& item) {
+CacheAllocator<CacheTrait>::tryEvictChainedItem(MMContainer& mmContainer, Item& item) {
   XDCHECK(item.isChainedItem());
 
   ChainedItem* candidate = &item.asChainedItem();
@@ -1417,9 +1435,11 @@ CacheAllocator<CacheTrait>::tryEvictChainedItem(Item& item) {
 
   // Ensure we have the correct parent and we're the only user of the
   // parent, then free it from access container. Otherwise, we abort
+  auto removed = mmContainer.remove(item);
+  XDCHECK(removed);
   XDCHECK_EQ(reinterpret_cast<uintptr_t>(&parent),
              reinterpret_cast<uintptr_t>(parentHandle.get()));
-  XDCHECK_EQ(1u, parent.getRefCount());
+  XDCHECK_EQ(2u, parent.getRefCount()); // XXX?
   XDCHECK(!parent.isInMMContainer());
   XDCHECK(!parent.isAccessible());
 
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -1228,13 +1228,15 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
     (*stats_.evictionAttempts)[pid][cid].inc();
 
     Item* candidate = itr.get();
+    mmContainer.remove(itr);
+    itr.destroy();
+
     // for chained items, the ownership of the parent can change. We try to
     // evict what we think as parent and see if the eviction of parent
     // recycles the child we intend to.
-    auto toReleaseHandle =
-        itr->isChainedItem()
-            ? advanceIteratorAndTryEvictChainedItem(itr)
-            : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
+    auto toReleaseHandle = candidate->isChainedItem()
+                               ? tryEvictChainedItem(*candidate)
+                               : tryEvictRegularItem(mmContainer, *candidate);
 
     if (toReleaseHandle) {
       if (toReleaseHandle->hasChainedItem()) {
@@ -1249,9 +1251,6 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
             AllocatorApiResult::EVICTED, toReleaseHandle->getSize(),
             toReleaseHandle->getConfiguredTTL().count());
       }
-      // Invalidate iterator since later on we may use this mmContainer
-      // again, which cannot be done unless we drop this iterator
-      itr.destroy();
 
       // we must be the last handle and for chained items, this will be
       // the parent.
@@ -1276,11 +1275,9 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
       }
     }
 
-    // If we destroyed the itr to possibly evict and failed, we restart
-    // from the beginning again
-    if (!itr) {
-      itr.resetToBegin();
-    }
+    // Insert item back to the mmContainer if eviction failed.
+    mmContainer.add(*candidate);
+    itr.resetToBegin();
   }
   return nullptr;
 }
@@ -1335,11 +1332,10 @@ bool CacheAllocator<CacheTrait>::shouldWriteToNvmCacheExclusive(
 
 template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::WriteHandle
-CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictRegularItem(
-    MMContainer& mmContainer, EvictionIterator& itr) {
+CacheAllocator<CacheTrait>::tryEvictRegularItem(MMContainer& mmContainer,
+                                                Item& item) {
   // we should flush this to nvmcache if it is not already present in nvmcache
   // and the item is not expired.
-  Item& item = *itr;
   const bool evictToNvmCache = shouldWriteToNvmCache(item);
 
   auto token = evictToNvmCache ? nvmCache_->createPutToken(item.getKey())
@@ -1347,7 +1343,6 @@ CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictRegularItem(
   // record the in-flight eviciton. If not, we move on to next item to avoid
   // stalling eviction.
   if (evictToNvmCache && !token.isValid()) {
-    ++itr;
     stats_.evictFailConcurrentFill.inc();
     return WriteHandle{};
   }
@@ -1359,12 +1354,10 @@ CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictRegularItem(
   auto evictHandle = accessContainer_->removeIf(item, &itemEvictionPredicate);
 
   if (!evictHandle) {
-    ++itr;
     stats_.evictFailAC.inc();
     return evictHandle;
   }
 
-  mmContainer.remove(itr);
   XDCHECK_EQ(reinterpret_cast<uintptr_t>(evictHandle.get()),
              reinterpret_cast<uintptr_t>(&item));
   XDCHECK(!evictHandle->isInMMContainer());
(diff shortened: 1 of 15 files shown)


cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2468,7 +2340,7 @@ void CacheAllocator<CacheTrait>::releaseSlabImpl(
     // Need to mark an item for release before proceeding
     // If we can't mark as moving, it means the item is already freed
     const bool isAlreadyFreed =
-        !markMovingForSlabRelease(releaseContext, alloc, throttler);
+        !markExclusiveForSlabRelease(releaseContext, alloc, throttler);
```
therealgymmy, line 2343: This is no longer safe. The original intent is that only a freed item cannot be marked as moving and no allocations can allocate a freed item once that slab was set to be released.
With this change, we can get false here if the item was already marked as exclusive (e.g. from an eviction thread). This means here we would treat that item as one already freed, which isn't the case.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2763,12 +2633,118 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
...
 
+template <typename CacheTrait>
+typename CacheAllocator<CacheTrait>::WriteHandle
+CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictRegularItem(
+    MMContainer& mmContainer, EvictionIterator& itr) {
+  // we should flush this to nvmcache if it is not already present in nvmcache
+  // and the item is not expired.
+  Item& item = *itr;
+  const bool evictToNvmCache = shouldWriteToNvmCache(item);
+
```
therealgymmy, line 2644: We still need this check that returns early if token is invalid. Otherwise we trigger the assert at L2677.

That said, we could just skip evicting to NvmCache on eviction if we couldn't acquire a valid PutToken (which can happen due to contention on the in-flight-puts queues). It may cause impact on hit rate. So it's best done as a separate change.

```
  // record the in-flight eviciton. If not, we move on to next item to avoid
  // stalling eviction.
  if (evictToNvmCache && !token.isValid()) {
    ++itr;
    stats_.evictFailConcurrentFill.inc();
    return WriteHandle{};
  }
```

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2763,12 +2633,118 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
...
+  // exclusive iterator so we can gain access to it when we're cleaning
+  // up the child items
+  itr.destroy();
+
+  // Ensure that there are no accessors after removing from the access
+  // container
+  XDCHECK(evictHandle->getRefCount() == 1);
+
+  if (evictToNvmCache && shouldWriteToNvmCacheExclusive(item)) {
+    XDCHECK(token.isValid());
```
therealgymmy, line 2679: This failed our internal tests. You can repro this by running this cachebench test config `cachelib/cachebench/test_configs/consistency/navy.json`. The issue is related to comments at L2644.

```
F1013 14:57:11.677257   633 CacheAllocator-inl.h:2677] Check failed: token.isValid() 
*** Aborted at 1665698231 (Unix time, try 'date -d @1665698231') ***
*** Signal 6 (SIGABRT) (0x6300000161) received by PID 353 (pthread TID 0x7f2645c00640) (linux TID 633) (maybe from PID 353, UID 99) (code: -6), stack trace: ***
    @ 000000000698b4be folly::symbolizer::(anonymous namespace)::signalHandler(int, siginfo_t*, void*)
                       ./folly/experimental/symbolizer/SignalHandler.cpp:449
    @ 0000000000000000 (unknown)
    @ 000000000009c9d3 __GI___pthread_kill
    @ 00000000000444ec __GI_raise
    @ 000000000002c432 __GI_abort
    @ 00000000069ac304 folly::LogCategory::admitMessage(folly::LogMessage const&) const
                       ./folly/logging/LogCategory.cpp:71
    @ 00000000069aa42d folly::LogStreamProcessor::logNow()
                       ./folly/logging/LogStreamProcessor.cpp:190
    @ 00000000069aa91c folly::LogStreamVoidify<true>::operator&(std::ostream&)
                       ./folly/logging/LogStreamProcessor.cpp:222
    @ 0000000005c608b6 facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait>::advanceIteratorAndTryEvictRegularItem(facebook::cachelib::MMLru::Container<facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait>, &facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait>::mmHook_>&, facebook::cachelib::MMLru::Container<facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait>, &facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait>::mmHook_>::Iterator&)
                       ./cachelib/allocator/CacheAllocator-inl.h:2677
                       -> ./cachelib/allocator/CacheAllocator.cpp
    @ 0000000005c5810d facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait>::findEviction(signed char, signed char)
                       ./cachelib/allocator/CacheAllocator-inl.h:1251
                       -> ./cachelib/allocator/CacheAllocator.cpp
    @ 0000000005c5897c facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait>::allocateChainedItemInternal(facebook::cachelib::detail::ReadHandleImpl<facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait> > const&, unsigned int)
                       ./cachelib/allocator/CacheAllocator-inl.h:409
                       -> ./cachelib/allocator/CacheAllocator.cpp
    @ 0000000005c8bee0 facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait>::allocateChainedItem(facebook::cachelib::detail::ReadHandleImpl<facebook::cachelib::CacheItem<facebook::cachelib::LruCacheTrait> > const&, unsigned int)
                       ./cachelib/allocator/CacheAllocator-inl.h:381
                       -> ./cachelib/allocator/CacheAllocator.cpp
    @ 00000000059c0e0a facebook::cachelib::cachebench::CacheStressor<facebook::cachelib::CacheAllocator<facebook::cachelib::LruCacheTrait> >::stressByDiscreteDistribution(facebook::cachelib::cachebench::ThroughputStats&)
                       ./cachelib/cachebench/cache/Cache-inl.h:355
                       -> ./cachelib/cachebench/runner/Stressor.cpp
    @ 00000000000df4e4 execute_native_thread_routine
    @ 000000000009ac0e start_thread
    @ 000000000012d1db __clone3

```

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2763,12 +2633,118 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
...
+  ++itr;
+
+  // The parent could change at any point through transferChain. However, if
+  // that happens, we would realize that the releaseBackToAllocator return
+  // kNotRecycled and we would try another chained item, leading to transient
+  // failure.
+  auto& parent = candidate->getParentItem(compressor_);
+
+  const bool evictToNvmCache = shouldWriteToNvmCache(parent);
+
```
therealgymmy, line 2701: same here we should return early if token isn't valid. Otherwise we can trigger the assertion at L2734.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2940,7 +2916,7 @@ bool CacheAllocator<CacheTrait>::markMovingForSlabRelease(
     // Since this callback is executed, the item is not yet freed
     itemFreed = false;
     Item* item = static_cast<Item*>(memory);
-    if (item->markMoving()) {
+    if (item->markExclusive()) {
```
therealgymmy, line 2919: This now returns false if the item was already marked as exclusive. This causes issue in slab release.

cachelib/allocator/CacheItem.h
```diff
@@ -349,20 +345,20 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
+   * An item can only be marked exclusive when `isInMMContainer` returns true.
    * This operation is atomic.
    *
-   * User can also query if an item "isOnlyMoving". This returns true only
-   * if the refcount is 0 and only the moving bit is set.
+   * User can also query if an item "isOnlyExclusive". This returns true only
+   * if the refcount is 0 and only the exclusive bit is set.
    *
-   * Unmarking moving does not depend on `isInMMContainer`
+   * Unmarking exclusive does not depend on `isInMMContainer`
```
therealgymmy, line 356: Can you add these to the comments

```
Unmarking exclusive will also return the refcount at the moment of unmarking.
```

cachelib/allocator/Refcount.h
```diff
@@ -247,18 +248,18 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+   * An item can only be marked exclusive when `isInMMContainer` returns true
+   * and the item is not yet marked as exclusive. This operation is atomic.
    *
-   * User can also query if an item "isOnlyMoving". This returns true only
-   * if the refcount is 0 and only the moving bit is set.
+   * User can also query if an item "isOnlyExclusive". This returns true only
+   * if the refcount is 0 and only the exclusive bit is set.
    *
-   * Unmarking moving does not depend on `isInMMContainer`
+   * Unmarking exclusive does not depend on `isInMMContainer`
```
therealgymmy, line 259: Same here. Can you add these to the comments.

```
Unmarking exclusive will also return the refcount at the moment of unmarking.
```

cachelib/allocator/Refcount.h
```diff
@@ -286,21 +288,23 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+  bool isExclusive() const noexcept {
+    return getRaw() & getAdminRef<kExclusive>();
   }
-  bool isMoving() const noexcept { return getRaw() & getAdminRef<kMoving>(); }
-  bool isOnlyMoving() const noexcept {
-    // An item is only moving when its refcount is zero and only the moving bit
-    // among all the control bits is set. This indicates an item is already on
-    // its way out of cache and does not need to be moved.
+  bool isOnlyExclusive() const noexcept {
+    // An item is only exclusive when its refcount is zero and only the
```
therealgymmy, line 299:
```
// An item is only exclusive when its refcount is zero and only the
// exclusive bit among all the control bits is set. This indicates an item
// is exclusive to the current thread. No other thread is allowed to
// do anything with it.
```

therealgymmy (changes requested):
The performance looks good. No noticeable diff in A/B tests. I left comments on the correctness of a couple of places.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1227,60 +1228,68 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+        stats_.evictFailAC.inc();
       }
     }
 
-    // If we destroyed the itr to possibly evict and failed, we restart
-    // from the beginning again
-    if (!itr) {
-      itr.resetToBegin();
-    }
+    itr.resetToBegin();
```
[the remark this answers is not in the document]
igchor: Right, done.

cachelib/allocator/CacheAllocator-inl.h
igchor: I think this cannot happen. If we fail to mark the item as exclusive inside `processAllocForRelease` we will just spin on the `while(true)` inside `markExclusiveForSlabRelease`

cachelib/allocator/CacheAllocator-inl.h
igchor: Got it, fixed.

cachelib/allocator/CacheAllocator-inl.h
igchor: Fixed.

cachelib/allocator/CacheAllocator-inl.h
igchor: Fixed

cachelib/allocator/CacheAllocator-inl.h
igchor: In case it returns false, we will just spin in the while loop until the item is either freed or we succeed in marking it as exclusive.

cachelib/allocator/CacheItem.h
igchor: Done.

cachelib/allocator/Refcount.h
igchor: Done.

igchor:
By the way, is there any fundamental reason for NvmCache::put to accept a handle instead of a raw pointer? I'm working on the next set of changes (based on this PR) and passing a raw ptr would be much simpler

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2962,7 +2953,7 @@ bool CacheAllocator<CacheTrait>::markMovingForSlabRelease(
     itemFreed = true;
 
     if (shutDownInProgress_) {
-      XDCHECK(!static_cast<Item*>(alloc)->isMoving());
+      XDCHECK(!static_cast<Item*>(alloc)->isExclusive());
```
therealgymmy, line 2956: Should we remove this assert? I think this can be true now if an item is marked as exclusive by eviction thread while the slab release gets to this line.

cachelib/allocator/CacheAllocator-inl.h
therealgymmy: yeah you're right. This will just repeat in the loop. I noticed one more thing L2956 that I think we should remove the assert there.

cachelib/allocator/CacheAllocator-inl.h
therealgymmy: ^ yup you're right.

therealgymmy:
@igchor: 

> By the way, is there any fundamental reason for NvmCache::put to accept a handle instead of a raw pointer? I'm working on the next set of changes (based on this PR) and passing a raw ptr would be much simpler

By raw pointer do you mean `Item&`? Why would it make your changes much simpler?

(It does seem possible to just pass in Item&)

cachelib/allocator/CacheAllocator-inl.h
igchor: Yes, you are right. Removed.

igchor:
> @igchor:
> 
> > By the way, is there any fundamental reason for NvmCache::put to accept a handle instead of a raw pointer? I'm working on the next set of changes (based on this PR) and passing a raw ptr would be much simpler
> 
> By raw pointer do you mean `Item&`? Why would it make your changes much simpler?
> 
> There is no strong reason to use WriteHandle&. Mostly the code was just written this way.

Ok, I overstated it a bit :) It will make it slightly easier because in the patch, in some places we do not acquire a handle to the item anymore (.e.g. in findEviction we only operate on pointers/references). Once this PR is merged I'll rebase my second patch and push it so you can take a look.

therealgymmy (approved):
LGTM

cachelib/allocator/CacheAllocator-inl.h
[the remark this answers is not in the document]
haowu14: In the version that's imported into Meta this is still not fixed and caused problems in production.

We would need to revert this and re-import it. We will try our best to prioritize the re-import. @igchor

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2763,12 +2633,133 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
...
+  auto& parent = candidate->getParentItem(compressor_);
+
+  const bool evictToNvmCache = shouldWriteToNvmCache(parent);
+
+  auto token = evictToNvmCache ? nvmCache_->createPutToken(parent.getKey())
+                               : typename NvmCacheT::PutToken{};
+
+  // if token is invalid, return. iterator is already advanced.
+  if (evictToNvmCache && !token.isValid()) {
+    ++itr;
```
haowu14, line 2715: We already advanced the iterator at line 2700 so we shouldn't do it again here.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1227,60 +1228,68 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+              ? advanceIteratorAndTryEvictChainedItem(itr)
+              : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
+      // destroy toReleseHandle. The item won't be release to allocator
+      // since we marked it as exclusive.
+    }
 
-    if (toReleaseHandle) {
-      if (toReleaseHandle->hasChainedItem()) {
+    const auto ref = candidate->unmarkExclusive();
+    if (ref == 0u) {
```
haowu14, line 1257: I think we still need to check if `toReleaseHandle` is valid. Not because we want to make sure its not being evicted by another thread, but to know whether we did actually removed `candidate` from the `advanceIteratorAndTryEvict*Item` function call.

If an item was not removed, we should skip the logic inside the if-else block and simply unmarkExclusive and move on. If we don't want to return toReleaseHandle because we don't want to increment the refcount, we should find another way to check if the iterator was advanced or not.

cachelib/allocator/CacheAllocator-inl.h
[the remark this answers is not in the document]
haowu14: This needs to be fixed.

haowu14 (commented):
Hi @igchor. This change caused an issue in production and we reverted it.

Once you get a chance to fix the comments, I'll import it again and run tests against the service it failed on.

cachelib/allocator/CacheAllocator-inl.h

cachelib/allocator/CacheAllocator-inl.h
igchor: Does it matter whether **we** did actually remove the `candidate`? If it was some other thread that did this and the ref is `0` after `unmarkExclsuive` we should still be able to recycle the item. `unmarkExclusive` in this case is like `decRef()` . For `ref == 0` we are the last owner but instead of returning the item to allocator we reuse it.

I even belive that we **have** to recycle the item (or release it back to allocator) in this case. Otherwise we could have a memory leak.

Is my thinking correct or do I miss something?

cachelib/allocator/CacheAllocator-inl.h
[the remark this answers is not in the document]
igchor: Done. Sorry about this, I think something went wrong when I did rebase to the latest main.

cachelib/allocator/CacheAllocator-inl.h
haowu14: I think you are right. `unmarkExclusive` has both admin and access refcount right? if so, you are correct that we would have to recycle as this is the last owner.

I think there is just one small problem now:

C1: removed item and ref == 0 (no problem. The golden path)
C2: removed item and ref != 0 (shouldn't happen. No access ref can be created after removed from AccessContainer)
C3: not removed and ref == 0 (no problem. If happened, should still recycle because another thread had removed the item from Access and MMContainer, but didn't recycle because current thread is holding exclusive bit)
C4: not removed and ref > 0. In this case, we would increment the eviction failure counters again in the else block, while we already incremented them inside `advanceIteratorAndTryEvict`.

I think we just need to fix C4 here.

igchor: Yes, `ref` will contain both access and admin bits.

You are also right about those 4 cases. Since C2 cannot happen, all the stats are always updated inside `advanceIteratorAndTryEvict*`. So, I just removed the code which updates stats outside of those functions and added an `assert` for C2.

haowu14:
@igchor Update: we will be testing this in a prod-like setup in the week of Nov 28.

haowu14:
Tested on Monday and landed just now. I'll move on to the next PR.

Issue #169, opened 2022-10-21
Test failures on CentOS 8


igchor:
**Describe the bug**
List of tests failing (some randomly):
* NvmCacheTest.EvictToNvmGet
```
I1021 16:07:49.816931 345143 Factory.cpp:289] Cache file: /tmp/nvmcache-cachedir/345143/navy size: 104857600 truncate: 0
E1021 16:07:49.816977 345143 Factory.cpp:305] Failed to open with o-direct, trying without. Error: open("/tmp/nvmcache-cachedir/345143/navy", 040102, 00666) failed: Invalid argument
I1021 16:07:49.818993 345143 NavySetup.cpp:89] metadataSize: 4194304 bigHashCacheOffset: 52428800 bigHashCacheSize: 52428800
I1021 16:07:49.819003 345143 NavySetup.cpp:121] blockcache: starting offset: 4194304, block cache size: 46137344
I1021 16:07:49.819077 345143 LruPolicy.cpp:37] LRU policy: expected 11 regions
I1021 16:07:49.823701 345143 RegionManager.cpp:50] 11 regions, 4194304 bytes each
I1021 16:07:49.823733 345143 Allocator.cpp:43] Enable priority-based allocation for Allocator. Number of priorities: 1
I1021 16:07:49.823761 345143 BlockCache.cpp:143] Block cache created
I1021 16:07:49.823954 345143 BigHash.cpp:106] BigHash created: buckets: 51200, bucket size: 1024, base offset: 52428800
I1021 16:07:49.823960 345143 BigHash.cpp:111] Reset BigHash
I1021 16:07:49.823976 345143 Driver.cpp:65] Max concurrent inserts: 1000000
I1021 16:07:49.823980 345143 Driver.cpp:66] Max parcel memory: 268435456
I1021 16:07:49.823986 345143 Driver.cpp:325] Reset Navy
I1021 16:07:49.823995 345143 BigHash.cpp:111] Reset BigHash
I1021 16:07:49.824006 345143 BlockCache.cpp:665] Reset block cache
I1021 16:07:49.860485 345143 BigHash.cpp:476] Flush big hash
I1021 16:07:49.860508 345143 BlockCache.cpp:659] Flush block cache
/opt/workspace/cachelib/allocator/nvmcache/tests/NvmCacheTests.cpp:189: Failure
Expected: (nullptr) != (hdl), actual: (nullptr) vs nullptr
```
* NvmCacheTest.EvictToNvmGetCheckCtime
```
[...]

PR #172, opened 2022-11-07
Critical section part 2: findEviction optimization and code unification

igchor:
Previous PR (https://github.com/facebook/CacheLib/pull/166) refactored code in findEviction so it would be easier to add support for multiple memory tiers.

Now, we have two possible approaches to move forward with adding multi-tier support:
1. Implement item movement between tiers as a separate code path in findEviction
2. Unify eviction and item movement (for single-tier, multi-tier and SlabRelease)

This PR implements the second approach. It unifies logic for item movement and eviction, allowing us to later reuse SlabRelease item movement logic for multi-tier implementation. This PR also enables the use of combined locking for MMContainer. It allows us to achieve much better performance in multi-tier scenario.

We have also observed 2X throughput improvement for leader benchmarks (on the upstream, single-tier version) with this PR. We have also seen an improvement in allocation latency. However, I have to admit that this patch is pretty complex. Please let me know if you're open to merging this or if you'd prefer not to modify so much code.

It would be great if you could validate those changes internally. I encountered one inconsistency when running consistency/navy.json benchmark but it's pretty rare (one per 1200M ops) and I'm not able to debug this. We'd like to validate those changes by running navy tests but unfortunately they are failing even on main branch: https://github.com/facebook/CacheLib/issues/169

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -463,9 +463,7 @@ void CacheAllocator<CacheTrait>::addChainedItem(WriteHandle& parent,
   // Parent will decrement the refcount upon release. Since this is an
   // internal refcount, we dont include it in active handle tracking.
   child->incRef();
-  auto ref = child->getRefCount();
-  // ref == 3 if child is moving
-  XDCHECK(ref == 2u || ref == 3);
+  XDCHECK_EQ(2u, child->getRefCount());
 
   invalidateNvm(*parent);
   if (auto eventTracker = getEventTracker()) {
@@ -555,10 +553,7 @@ void CacheAllocator<CacheTrait>::transferChainLocked(WriteHandle& parent,
   ChainedItem* curr = &headHandle->asChainedItem();
   const auto newParentPtr = compressor_.compress(newParent.get());
   while (curr) {
-    if (!curr->isMoving())
-      XDCHECK_EQ(curr == headHandle.get() ? 2u : 1u, curr->getRefCount());
-    else
-      XDCHECK_EQ(curr == headHandle.get() ? 3u : 2u, curr->getRefCount());
+    XDCHECK_EQ(curr == headHandle.get() ? 2u : 1u, curr->getRefCount());
     XDCHECK(curr->isInMMContainer());
     curr->changeKey(newParentPtr);
     curr = curr->getNext(compressor_);
@@ -654,7 +649,7 @@ CacheAllocator<CacheTrait>::replaceChainedItemLocked(Item& oldItem,
                                                      WriteHandle newItemHdl,
                                                      const Item& parent) {
   XDCHECK(newItemHdl != nullptr);
-  XDCHECK_GE(2u, oldItem.getRefCount());
+  XDCHECK_GE(1u, oldItem.getRefCount());
 
   // grab the handle to the old item so that we can return this. Also, we need
   // to drop the refcount the parent holds on oldItem by manually calling
@@ -1157,7 +1152,7 @@ bool CacheAllocator<CacheTrait>::moveRegularItem(Item& oldItem,
 
   // no one can add or remove chained items at this point
   if (oldItem.hasChainedItem()) {
-    auto oldItemHdl = WriteHandle{&oldItem, *this};
+    auto oldItemHdl = acquire(&oldItem);
     XDCHECK_EQ(1u, oldItemHdl->getRefCount()) << oldItemHdl->toString();
     XDCHECK(!newItemHdl->hasChainedItem()) << newItemHdl->toString();
     try {
@@ -1171,10 +1166,6 @@ bool CacheAllocator<CacheTrait>::moveRegularItem(Item& oldItem,
 
     XDCHECK(!oldItem.hasChainedItem());
     XDCHECK(newItemHdl->hasChainedItem());
-
-    // drop the handle, no need to decRef since we relied on
-    // item being moved
-    oldItemHdl.release();
   }
   newItemHdl.unmarkNascent();
   return true;
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -463,7 +463,9 @@ void CacheAllocator<CacheTrait>::addChainedItem(WriteHandle& parent,
   // Parent will decrement the refcount upon release. Since this is an
   // internal refcount, we dont include it in active handle tracking.
   child->incRef();
-  XDCHECK_EQ(2u, child->getRefCount());
+  auto ref = child->getRefCount();
+  // ref == 3 if child is moving
+  XDCHECK(ref == 2u || ref == 3);
 
   invalidateNvm(*parent);
   if (auto eventTracker = getEventTracker()) {
@@ -553,7 +555,10 @@ void CacheAllocator<CacheTrait>::transferChainLocked(WriteHandle& parent,
   ChainedItem* curr = &headHandle->asChainedItem();
   const auto newParentPtr = compressor_.compress(newParent.get());
   while (curr) {
-    XDCHECK_EQ(curr == headHandle.get() ? 2u : 1u, curr->getRefCount());
+    if (!curr->isMoving())
+      XDCHECK_EQ(curr == headHandle.get() ? 2u : 1u, curr->getRefCount());
+    else
+      XDCHECK_EQ(curr == headHandle.get() ? 3u : 2u, curr->getRefCount());
     XDCHECK(curr->isInMMContainer());
     curr->changeKey(newParentPtr);
     curr = curr->getNext(compressor_);
@@ -649,7 +654,7 @@ CacheAllocator<CacheTrait>::replaceChainedItemLocked(Item& oldItem,
                                                      WriteHandle newItemHdl,
                                                      const Item& parent) {
   XDCHECK(newItemHdl != nullptr);
-  XDCHECK_GE(1u, oldItem.getRefCount());
+  XDCHECK_GE(2u, oldItem.getRefCount());
 
   // grab the handle to the old item so that we can return this. Also, we need
   // to drop the refcount the parent holds on oldItem by manually calling
@@ -823,20 +828,23 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
 
       removeFromMMContainer(*head);
 
-      // If this chained item is marked as exclusive, we will not free it.
-      // We must capture the exclusive state before we do the decRef when
-      // we know the item must still be valid
-      const bool wasExclusive = head->isExclusive();
+      // If this chained item is marked as moving, we will not free it.
+      // We must capture the moving state before we do the decRef when
+      // we know the item must still be valid. Item cannot be marked as
+      // exclusive. Only parent can be marked as such and even parent needs
+      // to be unmark prior to calling releaseBackToAllocator.
+      const bool wasMoving = head->isMoving();
+      XDCHECK(!head->isExclusive());
 
       // Decref and check if we were the last reference. Now if the item
-      // was marked exclusive, after decRef, it will be free to be released
+      // was marked moving, after decRef, it will be free to be released
       // by slab release thread
       const auto childRef = head->decRef();
 
-      // If the item is already exclusive and we already decremented the
+      // If the item is already moving and we already decremented the
       // refcount, we don't need to free this item. We'll let the slab
       // release thread take care of that
-      if (!wasExclusive) {
+      if (!wasMoving) {
         if (childRef != 0) {
           throw std::runtime_error(folly::sformat(
               "chained item refcount is not zero. We cannot proceed! "
@@ -844,7 +852,7 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
               childRef, head->toString()));
         }
 
-        // Item is not exclusive and refcount is 0, we can proceed to
+        // Item is not moving and refcount is 0, we can proceed to
         // free it or recylce the memory
         if (head == toRecycle) {
           XDCHECK(ReleaseRes::kReleased != res);
@@ -872,9 +880,12 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
 }
 
 template <typename CacheTrait>
-void CacheAllocator<CacheTrait>::incRef(Item& it) {
-  it.incRef();
-  ++handleCount_.tlStats();
+bool CacheAllocator<CacheTrait>::incRef(Item& it) {
+  if (it.incRef()) {
+    ++handleCount_.tlStats();
+    return true;
+  }
+  return false;
 }
 
 template <typename CacheTrait>
@@ -894,8 +905,12 @@ CacheAllocator<CacheTrait>::acquire(Item* it) {
 
   SCOPE_FAIL { stats_.numRefcountOverflow.inc(); };
 
-  incRef(*it);
-  return WriteHandle{it, *this};
+  if (LIKELY(incRef(*it))) {
+    return WriteHandle{it, *this};
+  } else {
+    // item is being evicted
+    return WriteHandle{};
+  }
 }
 
 template <typename CacheTrait>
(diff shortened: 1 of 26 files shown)


cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -2638,267 +2659,102 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
...
+          !expectedParent.isAccessible() || !expectedParent.hasChainedItem()) {
+        continue;
+      }
 
-  // now that we are the only handle and we actually removed something from
-  // the RAM cache, we enqueue it to nvmcache.
-  if (evictToNvmCache && shouldWriteToNvmCacheExclusive(item)) {
-    nvmCache_->put(handle, std::move(token));
-  }
+      // search if the child is present in the chain
```
igchor, line 2690: I'm not really sure if this code for searching for a child is needed... I copied it from the original implementation of SlabRelease eviction logic, but it seems to me like checking if the item's pointer to the parent is valid should be enough.

cachelib/allocator/Refcount.h
```diff
@@ -132,17 +132,25 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+    Value bitMask = getAdminRef<kExclusive>();
     Value oldVal = __atomic_load_n(refPtr, __ATOMIC_RELAXED);
 
     while (true) {
+      const bool alreadyExclusive = oldVal & bitMask;
       const Value newCount = oldVal + static_cast<Value>(1);
       if (UNLIKELY((oldVal & kAccessRefMask) == (kAccessRefMask))) {
+        throw exception::RefcountOverflow("Refcount maxed out.");
+      }
+      if (alreadyExclusive && (oldVal & kAccessRefMask) == 0) {
```
haowu14, line 153: Why adding `(oldVal & kAccessRefMask) == 0`? I thought we would only check `alreadyExclusive` here .

cachelib/allocator/Refcount.h
```diff
@@ -192,8 +200,18 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
     }
   }
 
-  // Return refcount excluding control bits and flags
-  Value getAccessRef() const noexcept { return getRaw() & kAccessRefMask; }
+  // Return refcount excluding control bits and flags.
```
haowu14, line 203:
```suggestion
  // Return refcount excluding moving refcount, control bits and flags.
```

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
   }
 
   /**
-   * The following four functions are used to track whether or not
-   * an item is currently in the process of being moved. This happens during a
-   * slab rebalance or resize operation or during eviction.
+   * The following two functions corresond to whether or not an item is
+   * currently in the process of being evicted. When item is marked exclsuive
```
haowu14, line 268: What's your future plan for `markExclusive()`? Is it also used in a case other than eviction?
If not, I think we can rename this called markEviction() where markEviction means marking exclusive bit with access ref==0, isLinked and isAccessible.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+   * if the refcount is one and only the moving bit is set.
+   *
+   * Unmarking moving does not depend on `isInMMContainer`
+   */
+  bool markMoving() {
+    auto predicate = [](const Value curValue) {
+      Value conditionBitMask = getAdminRef<kLinked>();
+      const bool flagSet = curValue & conditionBitMask;
+      const bool alreadyExclusive = curValue & getAdminRef<kExclusive>();
+      const bool accessible = curValue & getAdminRef<kAccessible>();
```
haowu14, line 331: I don't see `accessible` is used here.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+      // user/
+      return (curValue + static_cast<Value>(1)) | getAdminRef<kExclusive>();
+    };
+
+    return markInternal(predicate, newValue);
+  }
+  Value unmarkMoving() noexcept {
+    XDCHECK(isMoving());
+    auto predicate = [](const Value curValue) {
+      XDCHECK((curValue & kAccessRefMask) != 0);
```
haowu14, line 355: Any reason for rechecking this here? `isMoving()` is checked outside.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+    auto newValue = [](const Value curValue) {
+      XDCHECK((curValue & kAccessRefMask) == 1);
+      return (curValue - static_cast<Value>(1));
+    };
+
+    return markInternal(predicate, newValue);
+  }
+
+  bool isOnlyMoving() const noexcept {
+    // An item is only moving when its refcount is one and only the moving bit
```
haowu14, line 386:
```suggestion
    // An item is only moving when its refcount is one and only the exclusive bit
```

cachelib/allocator/Refcount.h
```diff
@@ -132,17 +132,25 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
 
   // Bumps up the reference count only if the new count will be strictly less
-  // than or equal to the maxCount.
-  // @return true if refcount is bumped. false otherwise.
-  FOLLY_ALWAYS_INLINE bool incRef() noexcept {
+  // than or equal to the maxCount and the item is not exclusive
+  // @return true if refcount is bumped. false otherwise (if item is exclusive)
+  // @throw  exception::RefcountOverflow if new count would be greater than
+  // maxCount
+  FOLLY_ALWAYS_INLINE bool incRef() {
```
haowu14, line 139: Can you refactor this with `markInternal`?

cachelib/allocator/Refcount.h
```diff
@@ -370,6 +454,32 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
   }
 
  private:
+  template <typename P, typename F>
+  bool markInternal(P&& predicate, F&& newValueF) {
```
haowu14, line 458: Please add documentation.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1231,75 +1262,106 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
   // Keep searching for a candidate until we were able to evict it
   // or until the search limit has been exhausted
   unsigned int searchTries = 0;
-  auto itr = mmContainer.getEvictionIterator();
   while ((config_.evictionSearchTries == 0 ||
```
haowu14, line 1265: I'm a bit confused here. Are we trying to advance the iterator inside or outside the lambda function passed into `withEvictionIterator`?
My guess would be inside, because the second time you call `withEvictionIterator`, the iterator is a newly created one starting from the end of the queue. But this code seems to attempt to advance iterator both inside and outside that lambda.

haowu14 (changes requested):
I have only reviewed Refcount.h , NvmCache-inl.h and some of the eviction iterator change and have some questions.

cachelib/allocator/CacheAllocator-inl.h
igchor: Right, this is because in case of chained items, we might end up evicting the old parent, and the releaseBackToAllocator on line 1358 will return `kNotRecycled`. In that case, we'll just loop again and start looking for a new candidate (calling `withEvictionIterator` again). I did it this way to keep the behavior as close as possible to the original implementation.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+    auto newValue = [](const Value curValue) {
+      // Set exclusive flag and make the ref count non-zero (to distinguish
+      // from exclusive case). This extra ref will not be reported to the
+      // user/
+      return (curValue + static_cast<Value>(1)) | getAdminRef<kExclusive>();
+    };
+
+    return markInternal(predicate, newValue);
+  }
+  Value unmarkMoving() noexcept {
```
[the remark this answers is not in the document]
igchor: Done.

cachelib/allocator/Refcount.h
igchor: The idea is to distinguish between moving and forEviction. We only want to return false when the item is marked markedExclusive (marked for eviction).

I renamed the mark/unmark functions as you suggested so hopefully it makes more sense now.

cachelib/allocator/Refcount.h
igchor: Done.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+  /**
+   * The following functions correspond to whether or not an item is
+   * currently in the processed of being moved. When moving, ref count
+   * is always >= 1 and `kExclusive` bit is set.
+   *
+   * An item can only be marked moving when `isInMMContainer` returns true
+   * and item is not already exclusive nor moving.
+   *
+   * User can also query if an item "isOnlyMoving". This returns true only
+   * if the refcount is one and only the moving bit is set.
```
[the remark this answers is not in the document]
igchor: Right, done. I also fixed this in other places.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +264,131 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+  }
+
+  bool isOnlyMoving() const noexcept {
+    // An item is only moving when its refcount is one and only the moving bit
+    // among all the control bits is set. This indicates an item is already on
+    // its way out of cache and does not need to be moved.
     auto ref = getRefWithAccessAndAdmin();
-    bool anyOtherBitSet = ref & ~getAdminRef<kExclusive>();
-    if (anyOtherBitSet) {
+    Value valueWithoutMovingBit = ref & ~getAdminRef<kExclusive>();
```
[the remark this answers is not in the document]
igchor: Done.

cachelib/allocator/Refcount.h
igchor: Done.

cachelib/allocator/Refcount.h
igchor: Done.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1231,76 +1263,102 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+                  candidate)
+            mmContainer.remove(itr);
+          return;
+        }
 
-    Item* candidate =
-        toRecycle->isChainedItem()
-            ? &toRecycle->asChainedItem().getParentItem(compressor_)
-            : toRecycle;
+        if (candidate_->hasChainedItem()) {
```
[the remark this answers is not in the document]
haowu14: Actually this belongs to the `else` of `if (candidate_->markForEviction())`.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -823,28 +823,31 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
...
-      // If this chained item is marked as exclusive, we will not free it.
-      // We must capture the exclusive state before we do the decRef when
-      // we know the item must still be valid
-      const bool wasExclusive = head->isExclusive();
+      // If this chained item is marked as moving, we will not free it.
+      // We must capture the moving state before we do the decRef when
+      // we know the item must still be valid. Item cannot be marked as
+      // exclusive. Only parent can be marked as such and even parent needs
+      // to be unmark prior to calling releaseBackToAllocator.
+      const bool wasMoving = head->isMoving();
```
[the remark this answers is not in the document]
igchor: For chained items, we cannot mark the individual children as exclusive anymore - only moving. Since exclusive (markedForEviction) means that particular item is going to be evicted it can only be applied to the parent (it does not make sense to evict a single child).

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1231,76 +1263,102 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+        auto* candidate_ =
+            toRecycle_->isChainedItem()
+                ? &toRecycle_->asChainedItem().getParentItem(compressor_)
+                : toRecycle_;
+
+        token = createPutToken(*candidate_);
+
+        if (shouldWriteToNvmCache(*candidate_) && !token.isValid()) {
+          stats_.evictFailConcurrentFill.inc();
+        } else if (candidate_->markForEviction()) {
```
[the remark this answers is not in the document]
igchor: Yes, currently marking Item for eviction makes the item inaccessible for all reader (all find() calls will return NULL handle for item that is marked for eviction, as if it was already removed from Access Container). This means that we have to create the token earlier to guarantee consistency.

I might have got something wrong here, however. There is still (quite rare failure on consistency/navy.json benchmark).

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1231,76 +1263,102 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+          // nor holding a handle to that item
+
+          toRecycle = toRecycle_;
+          candidate = candidate_;
+
+          // Check if parent changed for chained items - if yes, we cannot
+          // remove the child from the mmContainer as we will not be evicting
+          // it. We could abort right here, but we need to cleanup in case
+          // unmarkForEviction() returns 0 - so just go through normal path.
+          if (!toRecycle_->isChainedItem() ||
```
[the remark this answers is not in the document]
igchor: Yes, that's correct. This is just an optimization for the common case.
If we succeed here, we don't have to take MMContainer lock for the second time in line 1337.

If the condition fails (the parent has changed), then - we could theoretically just move on to the next candidate (freeing the parent will no longer recycle the memory we want), but we need to do cleanup anyway.

igchor:
@haowu14 Thanks for the review!

Yes, splitting this PR is a good idea. I wanted to get your feedback on the overall approach first but if you belive this approach is correct I'm happy to work on splitting it.

I've already created an initial PR which adds combined locking support for MMContainer: https://github.com/facebook/CacheLib/pull/182

As for your comments:
1. `While moving, we call acquire(item) and we make assertions on the refcount. Those assertions need to be updated to reflect the fact that we now have refcount 1 under moving. (And maybe we don't need to acquire at all).`

I've actually made it so that `item.refCount()` will NOT return the extra ref which is used to signal that item is marked as moving. So, in case the item is marked, it's ref count can still be 0. 

2. `When unmakrMoving, do we always have refcount==1 and turning it into refcount==0 afterwards? For a successful move, we'd always have the above condition right?`

Yes, the internal refcount will be 1 and when we call `unmarkMoving` this will atomically clear the `kExclusive` flag and decrease refcount by 1. In case of successful move, the return value from `unmarkMoving` will be 0. In case of failure it can be != 0 (since item might not have been removed from AC or some reader increased the ref count).

`Coming back to the purpose of the PR on unifying movement and eviction. Can you provide more detail about what you mean on "unify"? I can see we refactored a few place such as moving creation of NVM put token, etc. When I see a change I'd like to associate it with how it may help in the future, but it's a bit hard. I can only verify the correctness but not the intention right now. Overall I think this diff is the right direction. And if you can break it down into multiple PRs it, we should be able to ship it gradually. I wonder what the other option looks like. Is it going to be less code change?`

In terms of benefits for the single-tier version, this PR introduces the possibility to use combined locking without impacting the eviction success rate which positively impacts performance in our benchmarks. (Using combined locking brings even more performance benefits for multi-tier implementation.)

The main intention under "unification" is to reuse as much code as possible for multi-tier implementation. Without the changes in this PR, we would have to implement two entirely separate code paths for findEviction (in the case of single-tier and multiple memory tiers). We would also need to duplicate or extend the Eviction and Movement logic for SlabRelease to support multiple tiers. Moreover, we recently added support for chained items to our fork that relies on this PR and I'm not sure if we would be able to implement that support as easily without this PR.

That said, some subset of changes from this PR is needed for multi-tier support in both options (e.g. distinguishing between markedForEviction and makedMoving). In a few days, we plan to clean up our git history and create a draft PR with multi-tier support so you can see what that looks like.

PR #182, opened 2022-12-13
Add combined locking support for MMContainer

igchor:
through withEvictionIterator function.

Also, expose the config option to enable and disable combined locking.
withEvictionIterator is implemented as an extra function, getEvictionIterator() is still there and it's behavior hasn't changed.

This is a subset of changes from: https://github.com/facebook/CacheLib/pull/172

### DIFF
--- a/cachelib/allocator/MM2Q.h
+++ b/cachelib/allocator/MM2Q.h
@@ -214,6 +214,50 @@ class MM2Q {
            size_t hotPercent,
            size_t coldPercent,
            uint32_t mmReconfigureInterval)
+        : Config(time,
+                 ratio,
+                 updateOnW,
+                 updateOnR,
+                 tryLockU,
+                 rebalanceOnRecordAccs,
+                 hotPercent,
+                 coldPercent,
+                 mmReconfigureInterval,
+                 false) {}
+
+    // @param time                    the LRU refresh time.
+    //                                An item will be promoted only once in each
+    //                                lru refresh time depite the
+    //                                number of accesses it gets.
+    // @param ratio                   the LRU refresh ratio. The ratio times the
+    //                                oldest element's lifetime in warm queue
+    //                                would be the minimum value of LRU refresh
+    //                                time.
+    // @param udpateOnW               whether to promote the item on write
+    // @param updateOnR               whether to promote the item on read
+    // @param tryLockU                whether to use a try lock when doing
+    // update.
+    // @param rebalanceOnRecordAccs   whether to do rebalance on access. If set
+    //                                to false, rebalance only happens when
+    //                                items are added or removed to the queue.
+    // @param hotPercent              percentage number for the size of the hot
+    //                                queue in the overall size.
+    // @param coldPercent             percentage number for the size of the cold
+    //                                queue in the overall size.
+    // @param mmReconfigureInterval   Time interval for recalculating lru
+    //                                refresh time according to the ratio.
+    // useCombinedLockForIterators    Whether to use combined locking for
+    //                                withEvictionIterator
+    Config(uint32_t time,
+           double ratio,
+           bool updateOnW,
+           bool updateOnR,
+           bool tryLockU,
+           bool rebalanceOnRecordAccs,
+           size_t hotPercent,
+           size_t coldPercent,
+           uint32_t mmReconfigureInterval,
+           bool useCombinedLockForIterators)
         : defaultLruRefreshTime(time),
           lruRefreshRatio(ratio),
           updateOnWrite(updateOnW),
@@ -223,7 +267,8 @@ class MM2Q {
           hotSizePercent(hotPercent),
           coldSizePercent(coldPercent),
           mmReconfigureIntervalSecs(
-              std::chrono::seconds(mmReconfigureInterval)) {
+              std::chrono::seconds(mmReconfigureInterval)),
+          useCombinedLockForIterators(useCombinedLockForIterators) {
       checkLruSizes();
     }
 
@@ -306,6 +351,9 @@ class MM2Q {
     // Minimum interval between reconfigurations. If 0, reconfigure is never
     // called.
     std::chrono::seconds mmReconfigureIntervalSecs{};
+
+    // Whether to use combined locking for withEvictionIterator.
+    bool useCombinedLockForIterators{false};
   };
 
   // The container object which can be used to keep track of objects of type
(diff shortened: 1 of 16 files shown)


haowu14:
I was reading your comment in https://github.com/facebook/CacheLib/pull/172#issuecomment-1349758485. In single-tier version, without the further change, would using combined_locking for iterator cause reduction in eviction success rate and why?

cachelib/allocator/MM2Q.h
```diff
@@ -68,6 +68,7 @@ class MM2Q {
   enum LruType { Warm, WarmTail, Hot, Cold, ColdTail, NumTypes };
 
   // Config class for MM2Q
+  // TODO: implement support for useCombinedLockForIterators
```
haowu14, line 71: Given the code you have, implementing this for 2Q is just adding the if statement you have for MMLRU right?

cachelib/allocator/tests/MMTypeTest.h
```diff
@@ -343,5 +348,26 @@ size_t MMTypeTest<MMType>::getListSize(const Container& c,
   }
   throw std::invalid_argument("LruType not existing");
 }
+
+template <typename MMType>
+void MMTypeTest<MMType>::verifyIterationVariants(Container& c) {
```
haowu14, line 353: Please add a comment on this function.

haowu14 (changes requested):
Overall this looks good. Once you address the qestions I will import it for internal review.

cachelib/allocator/MM2Q.h
igchor: Adding the if statement + extending the MM2Q config

igchor:
> I was reading your comment in [#172 (comment)](https://github.com/facebook/CacheLib/pull/172#issuecomment-1349758485). In single-tier version, without the further change, would using combined_locking for iterator cause reduction in eviction success rate and why?

There are two issues with using `withEvictionIterator` with current design:
1. Current implementation of `findEviction` makes it pretty tricky to use `withEvictionIterator` - the iterator is passed from one function to another.
2. The length of the critical section (under MMContainer lock) inside `findEviction`  is pretty big. Combined locking makes the p99 latency very high (especially for readers). 

The second issue is actually what started our efforts to optimize `findEviction` and shorten the Critical Section. We observed a very big performance drop in our initial multi-tier implementation (that followed the current `findEviction` logic). Our first attempt to shorten the Critical Section was this PR: https://github.com/facebook/CacheLib/pull/132 where we just wanted to mark the item as exclusive under the lock and do all other operations outside. The problem there was that the eviction success rate suffered leading us to https://github.com/facebook/CacheLib/pull/172 where performance is much better (for both single and multi-tier).

Lower eviction rate in PR 132 was due to higher chance of collision: after we marked the item and exited the critical section, someone else could have increased the ref count on that item causing us to fail. In the new design, this cannot happen because after item is marked, it's inaccessible by others.

cachelib/allocator/tests/MMTypeTest.h
igchor: Done.

cachelib/allocator/MM2Q.h
igchor: Done. I just realized that TinyLFU uses a spin lock, which does not support combined locking, so there is no option to enable it.

cachelib/allocator/MMLru.h
```diff
@@ -195,6 +200,9 @@ class MMLru {
     //     lruInsertionPointSpec = 2, we insert at a point 1/4th from tail
     uint8_t lruInsertionPointSpec{0};
 
+    // Whether to use combined locking for withEvictionIterator.
```
haowu14, line 203: Move this after `mmReconfigureIntervalSecs`. Meta internal build failed on the initialization order.

haowu14 (changes requested):

cachelib/allocator/MMLru.h
igchor: Done.

cachelib/allocator/MM2Q.h
```diff
@@ -213,7 +216,8 @@ class MM2Q {
            bool rebalanceOnRecordAccs,
            size_t hotPercent,
            size_t coldPercent,
-           uint32_t mmReconfigureInterval)
+           uint32_t mmReconfigureInterval,
+           bool useCombinedLockForIterators)
```
haowu14, line 220: Please make this optional. 
The way we typically do is create a new constructor with all the parameters while keeping the constructor of ```Config(uint32_t time,
           double ratio,
           bool updateOnW,
           bool updateOnR,
           bool tryLockU,
           bool rebalanceOnRecordAccs,
           size_t hotPercent,
           size_t coldPercent,
           uint32_t mmReconfigureInterval)```
And have it call the constructor with all the parameters with default value.

cachelib/allocator/tests/MMTypeTest.h
```diff
@@ -343,5 +348,28 @@ size_t MMTypeTest<MMType>::getListSize(const Container& c,
...
+// yields the same elements, in the same order.
+template <typename MMType>
+void MMTypeTest<MMType>::verifyIterationVariants(Container& c) {
+  std::vector<Node*> nodes;
+  for (auto iter = c.getEvictionIterator(); iter; ++iter) {
+    nodes.push_back(&(*iter));
+  }
+
+  size_t i = 0;
+  c.withEvictionIterator([&c, &nodes, &i](auto&& iter) {
```
haowu14, line 362: Don't have to capture `c` here since it is not used inside.

haowu14 (changes requested):

cachelib/allocator/MM2Q.h
igchor: Sure, done.

cachelib/allocator/tests/MMTypeTest.h
igchor: Done.

cachelib/allocator/MMTinyLFU.h
```diff
@@ -456,6 +444,40 @@ class MMTinyLFU {
       // Tiny and main cache iterators
       ListIterator tIter_;
       ListIterator mIter_;
+    };
+
+    class LockedIterator : public Iterator {
```
therealgymmy, line 449: @igchor: is this necessary? TinyLFU doesn't support combined locking. It seems like we can just rename Iterator to LockIterator and avoid introducing this new code here

igchor: Yes, you are right, I forgot to revert these changes after realizing TinyLFU does not support combined locking. Done.

PR #183, opened 2022-12-15
Introduce 'markedForEviction' state for the Item.

igchor:
It is similar to 'moving' but requires ref count to be 0.

An item which is marked for eviction causes all incRef() calls to that item to fail.

This will be used to ensure that once item is selected for eviction, no one can interfere and prevent the eviction from suceeding.

'markedForEviction' relies on the same 'exlusive' bit as the 'moving' state. To distinguish between those two states, 'moving' add 1 to the refCount. This is hidden from the user, so getRefCount() will not return that extra ref.

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -823,20 +823,21 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
 
       removeFromMMContainer(*head);
 
-      // If this chained item is marked as exclusive, we will not free it.
-      // We must capture the exclusive state before we do the decRef when
+      // If this chained item is marked as moving, we will not free it.
+      // We must capture the moving state before we do the decRef when
       // we know the item must still be valid
-      const bool wasExclusive = head->isExclusive();
+      const bool wasMoving = head->isMoving();
+      XDCHECK(!head->isMarkedForEviction());
 
       // Decref and check if we were the last reference. Now if the item
-      // was marked exclusive, after decRef, it will be free to be released
+      // was marked moving, after decRef, it will be free to be released
       // by slab release thread
       const auto childRef = head->decRef();
 
-      // If the item is already exclusive and we already decremented the
+      // If the item is already moving and we already decremented the
       // refcount, we don't need to free this item. We'll let the slab
       // release thread take care of that
-      if (!wasExclusive) {
+      if (!wasMoving) {
         if (childRef != 0) {
           throw std::runtime_error(folly::sformat(
               "chained item refcount is not zero. We cannot proceed! "
@@ -844,7 +845,7 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
               childRef, head->toString()));
         }
 
-        // Item is not exclusive and refcount is 0, we can proceed to
+        // Item is not moving and refcount is 0, we can proceed to
         // free it or recylce the memory
         if (head == toRecycle) {
           XDCHECK(ReleaseRes::kReleased != res);
@@ -872,9 +873,12 @@ CacheAllocator<CacheTrait>::releaseBackToAllocator(Item& it,
 }
 
 template <typename CacheTrait>
-void CacheAllocator<CacheTrait>::incRef(Item& it) {
-  it.incRef();
-  ++handleCount_.tlStats();
+bool CacheAllocator<CacheTrait>::incRef(Item& it) {
+  if (it.incRef()) {
+    ++handleCount_.tlStats();
+    return true;
+  }
+  return false;
 }
 
 template <typename CacheTrait>
@@ -894,8 +898,12 @@ CacheAllocator<CacheTrait>::acquire(Item* it) {
 
   SCOPE_FAIL { stats_.numRefcountOverflow.inc(); };
 
-  incRef(*it);
-  return WriteHandle{it, *this};
+  if (LIKELY(incRef(*it))) {
+    return WriteHandle{it, *this};
+  } else {
+    // item is being evicted
+    return WriteHandle{};
+  }
 }
 
 template <typename CacheTrait>
@@ -1170,7 +1178,7 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
 
   // This item has been unlinked from its parent and we're the only
   // owner of it, so we're done here
-  if (!oldItem.isInMMContainer() || oldItem.isOnlyExclusive()) {
+  if (!oldItem.isInMMContainer() || oldItem.isOnlyMoving()) {
     return false;
   }
 
@@ -1201,7 +1209,7 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
 
   // In case someone else had removed this chained item from its parent by now
   // So we check again to see if the it has been unlinked from its parent
-  if (!oldItem.isInMMContainer() || oldItem.isOnlyExclusive()) {
+  if (!oldItem.isInMMContainer() || oldItem.isOnlyMoving()) {
     return false;
   }
 
@@ -1217,7 +1225,7 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
   // parent's chain and the MMContainer.
   auto oldItemHandle =
       replaceChainedItemLocked(oldItem, std::move(newItemHdl), *parentHandle);
-  XDCHECK(oldItemHandle->isExclusive());
+  XDCHECK(oldItemHandle->isMoving());
   XDCHECK(!oldItemHandle->isInMMContainer());
 
   return true;
@@ -1246,7 +1254,7 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
             : toRecycle;
 
     // make sure no other thead is evicting the item
-    if (candidate->getRefCount() != 0 || !candidate->markExclusive()) {
+    if (candidate->getRefCount() != 0 || !candidate->markMoving()) {
       ++itr;
       continue;
     }
@@ -1261,11 +1269,11 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
               ? advanceIteratorAndTryEvictChainedItem(itr)
               : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
       evictionSuccessful = toReleaseHandle != nullptr;
-      // destroy toReleseHandle. The item won't be released to allocator
-      // since we marked it as exclusive.
+      // destroy toReleaseHandle. The item won't be released to allocator
+      // since we marked for eviction.
     }
 
-    const auto ref = candidate->unmarkExclusive();
+    const auto ref = candidate->unmarkMoving();
     if (ref == 0u) {
       // Invalidate iterator since later on we may use this mmContainer
       // again, which cannot be done unless we drop this iterator
@@ -2352,7 +2360,7 @@ void CacheAllocator<CacheTrait>::releaseSlabImpl(
     // Need to mark an item for release before proceeding
     // If we can't mark as moving, it means the item is already freed
     const bool isAlreadyFreed =
-        !markExclusiveForSlabRelease(releaseContext, alloc, throttler);
+        !markMovingForSlabRelease(releaseContext, alloc, throttler);
     if (isAlreadyFreed) {
       continue;
     }
@@ -2397,8 +2405,8 @@ bool CacheAllocator<CacheTrait>::moveForSlabRelease(
     stats_.numMoveAttempts.inc();
 
     // Nothing to move and the key is likely also bogus for chained items.
-    if (oldItem.isOnlyExclusive()) {
-      oldItem.unmarkExclusive();
+    if (oldItem.isOnlyMoving()) {
+      oldItem.unmarkMoving();
       const auto res =
           releaseBackToAllocator(oldItem, RemoveContext::kNormal, false);
       XDCHECK(res == ReleaseRes::kReleased);
@@ -2437,7 +2445,7 @@ bool CacheAllocator<CacheTrait>::moveForSlabRelease(
   // that's identical to this one to replace it. Here we just need to wait
   // until all users have dropped the item handles before we can proceed.
   startTime = util::getCurrentTimeSec();
-  while (!oldItem.isOnlyExclusive()) {
+  while (!oldItem.isOnlyMoving()) {
     throttleWith(throttler, [&] {
       XLOGF(WARN,
             "Spent {} seconds, slab release still waiting for refcount to "
@@ -2491,8 +2499,8 @@ CacheAllocator<CacheTrait>::allocateNewItemForOldItem(const Item& oldItem) {
       return {};
     }
 
-    // Set up the destination for the move. Since oldChainedItem would have
-    // the exclusive bit set, it won't be picked for eviction.
+    // Set up the destination for the move. Since oldChainedItem would be
+    // marked as moving, it won't be picked for eviction.
     auto newItemHdl =
         allocateChainedItemInternal(parentHandle, oldChainedItem.getSize());
     if (!newItemHdl) {
@@ -2544,7 +2552,7 @@ bool CacheAllocator<CacheTrait>::tryMovingForSlabRelease(
       // item is still valid.
       const std::string parentKey =
           oldItem.asChainedItem().getParentItem(compressor_).getKey().str();
-      if (oldItem.isOnlyExclusive()) {
+      if (oldItem.isOnlyMoving()) {
         // If chained item no longer has a refcount, its parent is already
         // being released, so we abort this try to moving.
         return false;
@@ -2574,10 +2582,10 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
     stats_.numEvictionAttempts.inc();
 
     // if the item is already in a state where only the exclusive bit is set,
-    // nothing needs to be done. We simply need to unmark exclusive bit and free
+    // nothing needs to be done. We simply need to call unmarkMoving and free
     // the item.
-    if (item.isOnlyExclusive()) {
-      item.unmarkExclusive();
+    if (item.isOnlyMoving()) {
+      item.unmarkMoving();
       const auto res =
           releaseBackToAllocator(item, RemoveContext::kNormal, false);
       XDCHECK(ReleaseRes::kReleased == res);
@@ -2608,7 +2616,7 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
       stats_.numEvictionSuccesses.inc();
 
       // we have the last handle. no longer need to hold on to the exclusive bit
-      item.unmarkExclusive();
+      item.unmarkMoving();
 
       // manually decrement the refcount to call releaseBackToAllocator
       const auto ref = decRef(*owningHandle);
@@ -2620,7 +2628,7 @@ void CacheAllocator<CacheTrait>::evictForSlabRelease(
     }
 
     if (shutDownInProgress_) {
-      item.unmarkExclusive();
+      item.unmarkMoving();
       allocator_->abortSlabRelease(ctx);
       throw exception::SlabReleaseAborted(
           folly::sformat("Slab Release aborted while trying to evict"
@@ -2766,9 +2774,9 @@ CacheAllocator<CacheTrait>::advanceIteratorAndTryEvictChainedItem(
 template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::WriteHandle
 CacheAllocator<CacheTrait>::evictNormalItemForSlabRelease(Item& item) {
-  XDCHECK(item.isExclusive());
+  XDCHECK(item.isMoving());
 
-  if (item.isOnlyExclusive()) {
+  if (item.isOnlyMoving()) {
     return WriteHandle{};
   }
 
@@ -2780,7 +2788,7 @@ CacheAllocator<CacheTrait>::evictNormalItemForSlabRelease(Item& item) {
 
   // We remove the item from both access and mm containers. It doesn't matter
   // if someone else calls remove on the item at this moment, the item cannot
-  // be freed as long as we have the exclusive bit set.
+  // be freed as long as it's marked for eviction.
   auto handle = accessContainer_->removeIf(item, std::move(predicate));
 
   if (!handle) {
@@ -2804,7 +2812,7 @@ CacheAllocator<CacheTrait>::evictNormalItemForSlabRelease(Item& item) {
 template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::WriteHandle
 CacheAllocator<CacheTrait>::evictChainedItemForSlabRelease(ChainedItem& child) {
-  XDCHECK(child.isExclusive());
+  XDCHECK(child.isMoving());
 
   // We have the child marked as moving, but dont know anything about the
   // state of the parent. Unlike the case of regular eviction where we are
@@ -2826,7 +2834,7 @@ CacheAllocator<CacheTrait>::evictChainedItemForSlabRelease(ChainedItem& child) {
   // check if the child is still in mmContainer and the expected parent is
   // valid under the chained item lock.
   if (expectedParent.getKey() != parentKey || !child.isInMMContainer() ||
-      child.isOnlyExclusive() ||
+      child.isOnlyMoving() ||
       &expectedParent != &child.getParentItem(compressor_) ||
       !expectedParent.isAccessible() || !expectedParent.hasChainedItem()) {
     return {};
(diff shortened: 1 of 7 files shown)


cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,143 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+   * User can also query if an item "isOnlyMoving". This returns true only
+   * if the refcount is one and only the exlusive bit is set.
+   *
+   * Unmarking moving does not depend on `isInMMContainer`
+   */
+  bool markMoving() {
+    auto predicate = [](const Value curValue) {
+      Value conditionBitMask = getAdminRef<kLinked>();
+      const bool flagSet = curValue & conditionBitMask;
+      const bool alreadyExclusive = curValue & getAdminRef<kExclusive>();
```
haowu14, line 316: Similar here. Please define `Value exclusiveBitMask = getAdminRef<kExclusive>();` outside the `predicate` and capture it into the lambda.

cachelib/allocator/CacheItem.h
```diff
@@ -344,23 +349,43 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
-   * An item can only be marked exclusive when `isInMMContainer` returns true.
+   * An item can only be marked exclusive when `isInMMContainer` returns true
+   * and item is not already exclusive nor moving and the ref count is 0.
    * This operation is atomic.
    *
-   * User can also query if an item "isOnlyExclusive". This returns true only
-   * if the refcount is 0 and only the exclusive bit is set.
-   *
-   * Unmarking exclusive does not depend on `isInMMContainer`.
+   * Unmarking exclusive does not depend on `isInMMContainer`
```
haowu14, line 358:
```suggestion
   * Unmarking for eviction does not depend on `isInMMContainer`
```

cachelib/allocator/CacheItem.h
```diff
@@ -344,23 +349,43 @@ class CACHELIB_PACKED_ATTR CacheItem {
 
   /**
    * The following two functions corresond to whether or not an item is
-   * currently in the process of being moved. This happens during a slab
-   * rebalance, eviction or resize operation.
+   * currently in the process of being evicted.
    *
-   * An item can only be marked exclusive when `isInMMContainer` returns true.
+   * An item can only be marked exclusive when `isInMMContainer` returns true
+   * and item is not already exclusive nor moving and the ref count is 0.
```
haowu14, line 355: Let's just say it "does not have exclusive bit set and access ref count is 0", which implies it is not moving.

cachelib/allocator/CacheItem.h
```diff
@@ -344,23 +349,43 @@ class CACHELIB_PACKED_ATTR CacheItem {
 
   /**
    * The following two functions corresond to whether or not an item is
-   * currently in the process of being moved. This happens during a slab
-   * rebalance, eviction or resize operation.
+   * currently in the process of being evicted.
    *
-   * An item can only be marked exclusive when `isInMMContainer` returns true.
+   * An item can only be marked exclusive when `isInMMContainer` returns true
```
haowu14, line 354:
```suggestion
   * An item can only be marked for eviction when `isInMMContainer` returns true
```

cachelib/allocator/CacheItem.h
```diff
@@ -344,23 +349,43 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
+  bool markForEviction() noexcept;
+  RefcountWithFlags::Value unmarkForEviction() noexcept;
+  bool isMarkedForEviction() const noexcept;
+
+  /**
+   * The following functions correspond to whether or not an item is
+   * currently in the processed of being moved. When moving, ref count
+   * is always >= 1.
+   *
+   * An item can only be marked moving when `isInMMContainer` returns true
```
haowu14, line 371: > item is not already exclusive nor moving.

Let's just talk about "does not have exclusive bit marked".

cachelib/allocator/CacheItem.h
```diff
@@ -344,23 +349,43 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
+   * Unmarking moving will also return the refcount at the moment of
+   * unmarking.
+   */
+  bool markMoving();
+  RefcountWithFlags::Value unmarkMoving() noexcept;
+  bool isMoving() const noexcept;
+  bool isOnlyMoving() const noexcept;
+
+  /** This function attempts to mark item as exclusive.
+   * Can only be called on the item that is moving.*/
```
haowu14, line 387:
```suggestion
  /** This function attempts to mark item for eviction.
   * Can only be called on the item that is moving and has the access refcount being 1.*/
```

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,143 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+  }
+
+  bool isMoving() const noexcept {
+    auto raw = getRaw();
+    return (raw & getAdminRef<kExclusive>()) && ((raw & kAccessRefMask) != 0);
+  }
+
+  /** This function attempts to mark item for eviction.
+   * Can only be called on the item that is moving.*/
+  bool markForEvictionWhenMoving() {
```
haowu14, line 365: It looks like an item can be linked (inMMContainer) but not accessible and be marked as moving. Then enter forEviction via this function.
Is this a problem? Should we check isLinked in this function as well?

haowu14 (changes requested):

cachelib/allocator/CacheItem.h
```diff
@@ -305,12 +305,17 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
-      return;
+  // Increments item's ref count
+  //
+  // @return true on success, failure if item is marked as exclusive
+  // @throw exception::RefcountOverflow on ref count overflow
+  FOLLY_ALWAYS_INLINE bool incRef() {
+    try {
+      return ref_.incRef();
+    } catch (exception::RefcountOverflow& e) {
+      throw exception::RefcountOverflow(
```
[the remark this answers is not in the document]
igchor: The reason I've added this rethrow is to keep the abstraction but if this does affect performance perhaps we could just build the message on RefCount.h level?

cachelib/allocator/CacheItem.h
```diff
@@ -305,12 +305,17 @@ class CACHELIB_PACKED_ATTR CacheItem {
...
   RefcountWithFlags::Value getRefCountAndFlagsRaw() const noexcept;
 
-  FOLLY_ALWAYS_INLINE void incRef() {
-    if (LIKELY(ref_.incRef())) {
-      return;
+  // Increments item's ref count
+  //
+  // @return true on success, failure if item is marked as exclusive
+  // @throw exception::RefcountOverflow on ref count overflow
+  FOLLY_ALWAYS_INLINE bool incRef() {
```
therealgymmy, line 312: Not a fan of changing `incRef` behavior to have 3 outcomes (success, fail due to exclusivity, refcount overflow). It complicates existing behavior.

My bigger concern is actually that I don't see a clear need for introducing `markedForEviction`.

Today the goal of preventing anyone else from accessing an item while it is being prepared for eviction is to remove it from accessContainer. Why do we need this change?

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,143 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
   }
 
   /**
-   * The following four functions are used to track whether or not
-   * an item is currently in the process of being moved. This happens during a
-   * slab rebalance or resize operation or during eviction.
+   * The following two functions correspond to whether or not an item is
```
therealgymmy, line 250: Marking an item for eviction will prevent anyone from incrementing refcounts.

Can we achieve the same by removing this item from mmContainer and accessContainer?

cachelib/allocator/Refcount.h
```diff
@@ -370,6 +449,39 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
  private:
+  /**
+   * Helper function to modify refCount_ atomically.
+   *
+   * If predicate(currentValue) is true, then it atomically assigns result
+   * of newValueF(currentValue) to refCount_ and returns true. Otherwise
+   * returns false and leaves refCount_ unmodified.
+   */
+  template <typename P, typename F>
+  bool atomicUpdateValue(P&& predicate, F&& newValueF) {
```
therealgymmy, line 460: nice. Thanks for abstracting this out.

therealgymmy (commented):
@igchor: can you help me understand why we need this change?

Today the goal of preventing anyone else from accessing an item while it is being prepared for eviction is to remove it from accessContainer. Is this insufficient?

My general impression reading the code is that we're changing the responsibility of managing an Item's state (and the right to access an item) to be exclusively within CacheItem logic itself. I don't understand why we're making this change. The change itself also doesn't seem to have reduced the complexity of eviction/moving management.

cachelib/allocator/CacheItem.h
igchor: > @igchor: can you help me understand why we need this change?
> 
> Today the goal of preventing anyone else from accessing an item while it is being prepared for eviction is to remove it from accessContainer. Is this insufficient?
> 
> My general impression reading the code is that we're changing the responsibility of managing an Item's state (and the right to access an item) to be exclusively within CacheItem logic itself. I don't understand why we're making this change. The change itself also doesn't seem to have reduced the complexity of eviction/moving management.

The main goal of this change is to improve performance.  This is a dependency of https://github.com/facebook/CacheLib/pull/172

As you noticed, marking an item for eviction is pretty much equivalent to removing it from `accessContainer`. However, the key difference is that `markForEviction` is lock-free. This is important because we would like to shorten MMContainer critical section (and possibly use combined locking), as I mentioned on https://github.com/facebook/CacheLib/pull/172

We could just call `markMoving`  instead of `markForEviction` inside MMContainer Critical Section. However, since this doesn't prevent users from obtaining new handles, the item marked as moving is not guaranteed to be evicted. This exact problem was causing a lower eviction success rate with my initial Critical Section Patch (which you found when running throughput/small_cache_chained_items.json)

igchor: Another reason is that we would like to leverage `moving` bit for data movement between memory tiers. The idea is to block readers (create a wait context) for items marked as `moving`. 

This means that we cannot call `markMoving` inside `findEviction` for items which are being evicted to NVM/deleted.  We need to have a different mechanism to synchronize evictions and SlabRelease: i.e. `markForEviction`

igchor: @therealgymmy here's a pseudocode of how we implemented support for multi-tier using `markMoving` and `markForEviction`:

```c++
//////////////////////////////////////////////////////////////////////////////
findEviction() {
	Item* candidate = NULL;
	mmContainer.withEvictionIterator(auto itr) {
		while (itr) {
			auto candidateMarked = 
				(lastTier && itr.get()->markForEviction()) ||
				(!lastTier && itr.get()->markMoving());
			if (candidateMarked) {
				// at this point, no one can obtain handle to the 'candidate'
				// if the item was marked for eviction, all readers will get NULL
				// if the item was marked moving, all readers will block until
				// wakeupReaders() is called
				candidate = itr.get();
				return;
			}
			
			itr++;
		}
	};
	
	if (!candidate) return NULL;
	
	if (!lastTier) {
		// tryEvictToNextMemoryTier can incur significant overhead - it might try to
		// evict the item on the layer below to make room for 'candidate'
		auto newHandle = tryEvictToNextMemoryTier(candidate);
		
		// item was successfuly moved between tiers
		if (newHandle) {
			// wakes up all readers waiting for the item
			wakeupReaders(candidate, newHandle);
		} else {
			// we failed to move the item to different tier, time to evict
			candidate->markExlusiveWhenMoving();
			removeFromAccessContainerAndevictItemToNVM(candidate);
			
			wakeupReaders(candidate, NULL);
			
		}
	} else {
		// since this is the last tier, candidate has already been marked for eviction,
		// evicition, will always suceed
		removeFromAccessContainerAndevictItemToNVM(candidate);
	}
	
	return candidate;
}

//////////////////////////////////////////////////////////////////////////////
// acquire is called when getting handle from AccessContainer (e.g. on find())
acquire(Item *it) {
	auto incRes = incRef(*it);
	if (incRes == incOK) {
		return WriteHandle(it);
	} else if (incRes == incFailedMarkedForEviction) {
		// item is being evicted
		return NULL;
	} else { // incRes = incFailedMoving
		// create wait context and wait until item is no longer moving
		return handleWithWaitContextForMovingItem(*it);
	}
}
```

cachelib/allocator/CacheItem.h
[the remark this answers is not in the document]
igchor: I've rewritten the comments.

igchor: Done.

cachelib/allocator/CacheItem.h
igchor: Done.

cachelib/allocator/CacheItem.h
igchor: Done.

cachelib/allocator/CacheItem.h
igchor: Done.

cachelib/allocator/CacheItem.h
igchor: I think that `item is moving` implies refcount is 1.

cachelib/allocator/Refcount.h
igchor: Yes, however, as I mentioned in the previous comment removing from accessContainer results in much higher contention (if done under mmContainer lock).

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,143 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
-    unsigned int nCASFailures = 0;
-    constexpr bool isWeak = false;
-    Value curValue = __atomic_load_n(refPtr, __ATOMIC_RELAXED);
-    while (true) {
+  bool markForEviction() noexcept {
+    auto predicate = [](const Value curValue) {
+      Value conditionBitMask = getAdminRef<kLinked>();
       const bool flagSet = curValue & conditionBitMask;
-      const bool alreadyExclusive = curValue & bitMask;
+      const bool alreadyExclusive = curValue & getAdminRef<kExclusive>();
```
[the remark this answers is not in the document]
igchor: Done. Is this what you had in mind?

cachelib/allocator/Refcount.h
igchor: Done.

cachelib/allocator/Refcount.h
igchor: Actually it does not matter whether the item `isLinked` at this point or not. However, I actually realised that `markForEviction` does not need to depend on `isAccessible` - this was just an optimization I implemented some time ago but it does not seem to matter anymore - I removed that check.

cachelib/allocator/CacheItem.h
therealgymmy: @igchor: thank you for clarifying with the pseudo-code. This now makes sense to me.

Re: markForEviction()

Can we set this bit to true even if there're concurrent accesses? The idea is to block future readers, and we can move this item to another queue for async eviction. (Obviously if 0 access, we can evict synchronously)

Not a suggestion for this diff but I am wondering if this will make the aggregate cost of evictions go down. Fewer skips are needed overtime.

igchor: That's a good idea. I think we have considered something like this in the past and even created some implementation but we didn't have a way to block readers back then - we just put the items from the head/tail of the list into the queue and relied on BG workers to move the items. Now, once we can block the readers, we should probably revisit this.

igchor: However, I think this will only work with `markMoving` and not `markForEviction` since the implementation uses `refCount` to distinguish between `forEviction` and `moving` states.

haowu14: @igchor  Do we want to change `acquire` now? After this PR merges, when the item has exclusive bit set, `acquire` will create a handle without incrementing refcount.

cachelib/allocator/Refcount.h
haowu14: I'm thinking we should do it the other way round: both makrForEviction and markMoving should check isAccessible.
Today, moving need the item to be in access container to succeed. (see moveRegularItem in CacheAllocator-inl.h).

haowu14 (changes requested):

cachelib/allocator/CacheItem.h
igchor: Yes, you're right, it's better to change it now to have consistent behavior. Done.

cachelib/allocator/Refcount.h
igchor: The problem is that we cannot require `isAccessible` for `markMoving` because `markMoving` is used also for SlabRelease and it should work for items not in the Access Container as well.

I can restore the check for `markForEviction` but as I said it's only an optimization. It's true that we require the item to be `accessible` in `moveRegularItem` but item can become inaccessible anytime after being marked as moving anyway (for example by concurrent `insertOrReplace`).

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -894,7 +898,12 @@ CacheAllocator<CacheTrait>::acquire(Item* it) {
...
   SCOPE_FAIL { stats_.numRefcountOverflow.inc(); };
 
-  incRef(*it);
+  if (LIKELY(incRef(*it))) {
+    return WriteHandle{it, *this};
+  } else {
+    // item is being evicted
+    return WriteHandle{};
+  }
   return WriteHandle{it, *this};
```
haowu14, line 907: Remove this line.

cachelib/allocator/Refcount.h
haowu14: You are right. We don't require accessible today in markExcluive (Basically mark moving).

cachelib/allocator/CacheItem.h
haowu14: In this PR when findEviction, we mark an item as moving (exclusive).
Then in `advanceIteratorAndTryEvictRegularItem` we call accessContainer_.removeIf, it will fail because it fails to increment the refcount (exclusive). It doesn't seem right?

haowu14 (changes requested):
Sending back to you for questions. Maybe I miss something.
Thanks for addressing my previous comments.

cachelib/allocator/CacheAllocator-inl.h

cachelib/allocator/CacheItem.h
igchor: Right now only `markForEviction` causes `incRef` to return NULL handle. Users can still obtain handles to items that are `markedMoving` (and we are using `markMoving` in advanceIteratorAndTryEvictRegularItem).

In the next step we want to change the behavior of `advanceIteratorAndTryEvictRegularItem` to actually rely on `markForEviction` but this will require some additional changes so I have not included them here. You can find them on https://github.com/facebook/CacheLib/pull/172

haowu14: Got it now. Thanks for explaining.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,155 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+
+    return retValue & kRefMask;
+  }
+
+  bool isMoving() const noexcept {
+    auto raw = getRaw();
+    return (raw & getAdminRef<kExclusive>()) && ((raw & kAccessRefMask) != 0);
+  }
+
+  /** This function attempts to mark item for eviction.
```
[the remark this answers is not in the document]
therealgymmy: Separately, @igchor: I am wondering if the changes in this PR are introducing a form of "upgrade" and "write" locks to our cache item.

The analogy I have in mind:
1. Multiple readers can access an item. (No exclusive bit set. IncRef() act as the way to getting hold of the read-lock, which is released when ItemHandle goes out of scope)
2. Multiple readers and one upgrade can co-exist. (Exclusive bit set. Refcount > 1)
3. Only one upgrade lock can be held at a time. (First thread setting exclusive bit wins)
4. Upgrade lock can be upgraded to a write lock but only without any concurrent readers. (Exclusive bit set. Refcount == 1 -> Refcount == 0)
5. Write lock can be held without an upgrade lock in place as long as no concurrent readers.
6. Readers are blocked when write lock is held. (Refcount fails when exclusive bit is set)

Once I start thinking in terms of read/upgrade/write, I'm starting to appreciate the simplicity brought by this PR much more.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,155 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
-   * Unmarking exclusive does not depend on `isInMMContainer`.
-   * Unmarking exclusive will also return the refcount at the moment of
-   * unmarking.
+   * Unmarking for eviction clears the `kExclusive` bit. `unamrkForEviction`
+   * does not depend on `isInMMContainer` nor `isAccessible`
    */
-  bool markExclusive() noexcept {
-    Value bitMask = getAdminRef<kExclusive>();
-    Value conditionBitMask = getAdminRef<kLinked>();
+  bool markForEviction() noexcept {
```
therealgymmy, line 267: TODO: read over this logic

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,155 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
-    Value* const refPtr = &refCount_;
-    unsigned int nCASFailures = 0;
-    constexpr bool isWeak = false;
-    Value curValue = __atomic_load_n(refPtr, __ATOMIC_RELAXED);
-    while (true) {
-      const bool flagSet = curValue & conditionBitMask;
-      const bool alreadyExclusive = curValue & bitMask;
-      if (!flagSet || alreadyExclusive) {
+    auto predicate = [linkedBitMask, exclusiveBitMask](const Value curValue) {
+      const bool linked = curValue & linkedBitMask;
```
therealgymmy, line 272: nitpick again: `notLinked = !(curValue & linkedBitMask)`

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,155 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+   * currently in the process of being evicted.
+   *
+   * An item that is marked for eviction prevents from obtaining a handle to
+   * the item (incRef() will return false). This guarantees that eviction of
+   * marked item will always suceed.
    *
-   * An item can only be marked exclusive when `isInMMContainer` returns true
-   * and the item is not yet marked as exclusive. This operation is atomic.
+   * An item can only be marked for eviction when `isInMMContainer` and
+   * `isAccessible` return true and item does not have `kExclusive` bit set
```
therealgymmy, line 258: is this still true? We don't check for `isAccessible` here.

Separately, curious if we can remove the condition `isInMMContainer == true` down the road?

therealgymmy (commented):
@igchor: I made another pass. LGTM overall. Comments inline.

cachelib/allocator/Refcount.h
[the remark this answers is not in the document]
igchor: Yes, that's a good analogy. The logic changes slightly once we introduce WaitContext for moving items (in that case, 'readers' wait for 'upgrade' to finish it's work), but the reasoning for 'upgrading' still holds.

We even had an idea of using the 'moving' bit and Wait Context to replace some existing synchronization for Chained Items, but we haven't done any implementation yet.

cachelib/allocator/Refcount.h
```diff
@@ -246,65 +247,155 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
-  bool isOnlyExclusive() const noexcept {
-    // An item is only exclusive when its refcount is zero and only the
-    // exclusive bit among all the control bits is set. This indicates an item
-    // is exclusive to the current thread. No other thread is allowed to
-    // do anything with it.
+
+  bool isOnlyMoving() const noexcept {
+    // An item is only moving when its refcount is one and only the exclusive
+    // bit among all the control bits is set. This indicates an item is already
+    // on its way out of cache and does not need to be moved.
```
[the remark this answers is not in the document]
igchor: Right, the last part is not true - I removed it. We should be able to extend this description in later PR (when we introduce Wait Context for moving items).

cachelib/allocator/Refcount.h
```diff
@@ -370,6 +461,39 @@ class FOLLY_PACK_ATTR RefcountWithFlags {
...
+                                      __ATOMIC_ACQ_REL, __ATOMIC_RELAXED)) {
+        return true;
+      }
+
+      if ((++nCASFailures % 4) == 0) {
+        // this pause takes up to 40 clock cycles on intel and the lock cmpxchgl
+        // above should take about 100 clock cycles. we pause once every 400
+        // cycles or so if we are extremely unlucky.
+        folly::asm_volatile_pause();
+      }
```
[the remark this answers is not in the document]
igchor: We actually had some discussion and did some measurements of this internally. I'll try to get the data and see if it makes sense to change this.

haowu14:
@igchor  We are ready for the next PR! Thanks for working on this.

igchor:
@haowu14 thanks! The next PR is coming soon!

PR #188, opened 2023-01-17
Compressed pointer addressing in single and 2-tier mode

guptask:
In this change, compressed pointer stores the tier index, slab index and alloc index. With slab worth 22 bits of data, if we have the min allocation size as 64 bytes, that requires 16 bits for storing the alloc index.  The tier id occupies the 32nd bit only since its value cannot exceed 2. This leaves the remaining 15 bits  for the  slab  index. Hence we can index 128 GiB of memory per tier in multi-tier configuration or index 256 GiB in single-tier configuration.
Backward compatibility has been ensured in this change. Till the multi-tier code is up or review, modular change such as compressed pointer ensures backward compatibility. This code will undergo some change when the multi-tier code is ready for review.

### DIFF
--- a/cachelib/allocator/CCacheAllocator.cpp
+++ b/cachelib/allocator/CCacheAllocator.cpp
@@ -36,7 +36,9 @@ CCacheAllocator::CCacheAllocator(MemoryAllocator& allocator,
       currentChunksIndex_(0) {
   auto& currentChunks = chunks_[currentChunksIndex_];
   for (auto chunk : *object.chunks()) {
-    currentChunks.push_back(allocator_.unCompress(CompressedPtr(chunk)));
+    // TODO : pass multi-tier flag when compact cache supports multi-tier config
+    currentChunks.push_back(
+        allocator_.unCompress(CompressedPtr(chunk), false /* isMultiTier */));
   }
 }
 
@@ -97,7 +99,9 @@ CCacheAllocator::SerializationType CCacheAllocator::saveState() {
 
   std::lock_guard<std::mutex> guard(resizeLock_);
   for (auto chunk : getCurrentChunks()) {
-    object.chunks()->push_back(allocator_.compress(chunk).saveState());
+    // TODO : pass multi-tier flag when compact cache supports multi-tier config
+    object.chunks()->push_back(
+        allocator_.compress(chunk, false /* isMultiTier */).saveState());
   }
   return object;
 }
--- a/cachelib/allocator/memory/CompressedPtr.h
+++ b/cachelib/allocator/memory/CompressedPtr.h
@@ -27,18 +27,29 @@ namespace cachelib {
 
 class SlabAllocator;
 
-// the following are for pointer compression for the memory allocator.  We
-// compress pointers by storing the slab index and the alloc index of the
-// allocation inside the slab. With slab worth kNumSlabBits of data, if we
-// have the min allocation size as 64 bytes, that requires kNumSlabBits - 6
-// bits for storing the alloc index. This leaves the remaining (32 -
-// (kNumSlabBits - 6)) bits for the slab index.  Hence we can index 256 GiB
-// of memory in slabs and index anything more than 64 byte allocations inside
-// the slab using a 32 bit representation.
-//
 // This CompressedPtr makes decompression fast by staying away from division and
 // modulo arithmetic and doing those during the compression time. We most often
-// decompress a CompressedPtr than compress a pointer while creating one.
+// decompress a CompressedPtr than compress a pointer while creating one. This
+// is used for pointer compression by the memory allocator.
+
+// We compress pointers by storing the tier index, slab index and alloc index of
+// the allocation inside the slab.
+
+// In original design (without memory tiers):
+// Each slab addresses 22 bits of allocations (kNumSlabBits). This is split into
+// allocation index and allocation size. If we have the min allocation size of
+// 64 bytes (kMinAllocPower = 6 bits), remaining kNumSlabBits(22) -
+// kMinAllocPower(6) = 16 bits for storing the alloc index. This leaves the
+// remaining 32 - (kNumSlabBits - kMinAllocPower) = 16 bits  for the  slab
+// index. Hence we can index 256 GiB of memory.
+
+// In multi-tier design:
+// kNumSlabIds and kMinAllocPower remains unchanged. The tier id occupies the
+// 32nd bit only since its value cannot exceed kMaxTiers(2). This leaves the
+// remaining 32 - (kNumSlabBits - kMinAllocPower) - 1 bit for tier id = 15 bits
+// for the slab index. Hence we can index 128 GiB of memory per tier in
+// multi-tier configuration.
+
 class CACHELIB_PACKED_ATTR CompressedPtr {
  public:
   using PtrType = uint32_t;
@@ -62,9 +73,10 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
     return static_cast<uint32_t>(1) << (Slab::kMinAllocPower);
   }
 
-  // maximum adressable memory for pointer compression to work.
+  // maximum addressable memory for pointer compression to work.
   static constexpr size_t getMaxAddressableSize() noexcept {
-    return static_cast<size_t>(1) << (kNumSlabIdxBits + Slab::kNumSlabBits);
+    return static_cast<size_t>(1)
+           << (numSlabIdxBits(false) + Slab::kNumSlabBits);
   }
 
   // default construct to nullptr.
@@ -89,8 +101,11 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
   PtrType ptr_{kNull};
 
   // create a compressed pointer for a valid memory allocation.
-  CompressedPtr(uint32_t slabIdx, uint32_t allocIdx)
-      : ptr_(compress(slabIdx, allocIdx)) {}
+  CompressedPtr(uint32_t slabIdx,
+                uint32_t allocIdx,
+                bool isMultiTiered,
+                TierId tid = 0)
+      : ptr_(compress(slabIdx, allocIdx, isMultiTiered, tid)) {}
 
   constexpr explicit CompressedPtr(PtrType ptr) noexcept : ptr_{ptr} {}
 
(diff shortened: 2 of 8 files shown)


cachelib/allocator/CCacheAllocator.cpp
```diff
@@ -36,7 +36,8 @@ CCacheAllocator::CCacheAllocator(MemoryAllocator& allocator,
       currentChunksIndex_(0) {
   auto& currentChunks = chunks_[currentChunksIndex_];
   for (auto chunk : *object.chunks()) {
-    currentChunks.push_back(allocator_.unCompress(CompressedPtr(chunk)));
+    // TODO : pass multi-tier flag when compact cache supports multi-tier config
+    currentChunks.push_back(allocator_.unCompress(CompressedPtr(chunk), false));
```
therealgymmy, line 40: just a nit: `false /* isMultiTier */`

Can you also annotate in other places we pass in the boolean for compress and uncompress calls?

cachelib/allocator/memory/CompressedPtr.h
```diff
@@ -27,18 +27,19 @@ namespace cachelib {
...
+// is used for pointer compression by the memory allocator.
+
+// We compress pointers by storing the tier index, slab index and alloc index of
+// the allocation inside the slab. With slab worth kNumSlabBits (22 bits) of
+// data, if we have the min allocation size as 64 bytes, that requires
+// kNumSlabBits - 6 = 16 bits for storing the alloc index.  The tier id occupies
+// the 32nd bit only since its value cannot exceed kMaxTiers (2). This leaves
+// the remaining (32 - (kNumSlabBits - 6) - 1 bit for tier id) =  15 bits  for
+// the  slab  index. Hence we can index 128 GiB of memory per tier in multi-tier
+// configuration or index 256 GiB in single-tier configuration.
```
therealgymmy, line 42: nit: can you add more details to the pointer math for both single-tier and multi-tier config?

e.g.

```
// Single-tier Pointer Compression
// With slab worth kNumSlabBits of data, if we
// have the min allocation size as 64 bytes, that requires kNumSlabBits - 6
// bits for storing the alloc index. This leaves the remaining (32 -
// (kNumSlabBits - 6)) bits for the slab index.  Hence we can index 256 GiB
// of memory in slabs and index anything more than 64 byte allocations inside
// the slab using a 32 bit representation.

// Multi-tier Point Compression
// With slab worth kNumSlabBits (22 bits) of
// data, if we have the min allocation size as 64 bytes, that requires
// kNumSlabBits - 6 = 16 bits for storing the alloc index.  The tier id occupies
// the 32nd bit only since its value cannot exceed kMaxTiers (2). This leaves
// the remaining (32 - (kNumSlabBits - 6) - 1 bit for tier id) =  15 bits  for
// the  slab  index. Hence we can index 128 GiB of memory per tier.
```

cachelib/allocator/memory/CompressedPtr.h
```diff
@@ -100,31 +105,56 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
...
+  // Use 32nd bit position for TierId
+  static constexpr unsigned int kNumTierIdxOffset = 31;
+
   static constexpr PtrType kAllocIdxMask = ((PtrType)1 << kNumAllocIdxBits) - 1;
 
+  // kNumTierIdxBits most significant bits
+  static constexpr PtrType kTierIdxMask = (PtrType)1 << kNumTierIdxOffset;
+
   // Number of bits for the slab index. This will be the top 16 bits of the
   // compressed ptr.
```
therealgymmy, line 117: Update comments here. With multi-tier, the slab index is top 15 bits.

cachelib/allocator/memory/SlabAllocator.h
```diff
@@ -246,19 +247,20 @@ class SlabAllocator {
...
+                                   bool isMultiTiered) const {
     if (ptr.isNull()) {
       return nullptr;
     }
 
-    const SlabIdx slabIndex = ptr.getSlabIdx();
-    const uint32_t allocIdx = ptr.getAllocIdx();
+    const SlabIdx slabIndex = ptr.getSlabIdx(isMultiTiered);
+    const uint32_t allocIdx = ptr.getAllocIdx(isMultiTiered);
     const Slab* slab = &slabMemoryStart_[slabIndex];
```
therealgymmy, line 264: Can you add a comment/TODO here just to clarify multi-tier has no effect here in accessing slab memory since we haven't incorporated the actual multi-tier logic yet.

therealgymmy (commented):
looks good. I'm not expecting a perf change since the boolean flag is either always true or false at runtime, but will run a sanity test to be sure.

cachelib/allocator/memory/CompressedPtr.h
```diff
@@ -89,8 +91,11 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
   PtrType ptr_{kNull};
 
   // create a compressed pointer for a valid memory allocation.
-  CompressedPtr(uint32_t slabIdx, uint32_t allocIdx)
-      : ptr_(compress(slabIdx, allocIdx)) {}
+  CompressedPtr(uint32_t slabIdx,
+                uint32_t allocIdx,
+                bool isMultiTiered,
```
jaesoo-fb, line 96: `isMultiTiered` looks redundant. We are storing the tid at the MSB of the compressed ptr. In this case, we can just interpret `isMultiTiered = false` if `tid == 0`. Additional logic in `CompressedPtr` can be simplified a lot in this way.

cachelib/allocator/memory/CompressedPtr.h
```diff
@@ -100,31 +105,56 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
...
+    auto noTierIdPtr = isMultiTiered ? ptr_ & ~kTierIdxMask : ptr_;
+    return static_cast<uint32_t>(noTierIdPtr >> kNumAllocIdxBits);
   }
 
   // Get the allocation index of the compressed ptr
-  uint32_t getAllocIdx() const noexcept {
+  uint32_t getAllocIdx(bool isMultiTiered) const noexcept {
+    XDCHECK(!isNull());
+    auto noTierIdPtr = isMultiTiered ? ptr_ & ~kTierIdxMask : ptr_;
+    return static_cast<uint32_t>(noTierIdPtr & kAllocIdxMask);
```
jaesoo-fb, line 148: This change is needed. MSB 16 bits of `kAllocIdxMask` should be 0. For `isMultiTiered` would not be used, but looks OK if you want to have for consistency?

cachelib/allocator/memory/CompressedPtr.h
```diff
@@ -100,31 +105,56 @@ class CACHELIB_PACKED_ATTR CompressedPtr {
...
+  }
+
+  uint32_t getTierId(bool isMultiTiered) const noexcept {
     XDCHECK(!isNull());
-    return static_cast<uint32_t>(ptr_ & kAllocIdxMask);
+    return isMultiTiered ? static_cast<uint32_t>(ptr_ >> kNumTierIdxOffset) : 0;
+  }
+
+  void setTierId(TierId tid) noexcept {
+    ptr_ += static_cast<uint32_t>(tid) << kNumTierIdxOffset;
```
jaesoo-fb, line 157: What is the used of this? Is this needed for succeeding changes?

cachelib/allocator/memory/SlabAllocator.h
```diff
@@ -246,19 +247,20 @@ class SlabAllocator {
         static_cast<uint32_t>(reinterpret_cast<const uint8_t*>(ptr) -
                               reinterpret_cast<const uint8_t*>(slab)) /
         allocSize;
-    return CompressedPtr{slabIndex, allocIdx};
+    return CompressedPtr{slabIndex, allocIdx, isMultiTiered};
```
jaesoo-fb, line 250: So, tier id is not set here. How are you going to provide the tier id?

I think you are going to have a separate `MemoryAllocator` per tier, so separate `SlabAllocator`. If so, I think the tier ID could be a member variable of the `SlabAllocator`? 

What about the `PtrCompressor` then? Would `CacheAllocator` create separate `PtrCompressor` per `MemoryAllocator`?

jaesoo-fb (commented):
It is hard to figure out if this change makes sense without the succeeding changes. Why don't you include other commits in the stack in the pull request?

cachelib/allocator/CCacheAllocator.cpp
guptask: done

cachelib/allocator/memory/SlabAllocator.h
guptask: done

cachelib/allocator/memory/CompressedPtr.h
guptask: done

cachelib/allocator/memory/CompressedPtr.h
guptask: done

guptask:
> It is hard to figure out if this change makes sense without the succeeding changes. Why don't you include other commits in the stack in the pull request?

Decision was made to upstream the multi-tier changes in phases. This upstream plan has been discussed with Meta as well. Please refer to https://github.com/intel/CacheLib/pull/56 for the phased plan.

cachelib/allocator/memory/CompressedPtr.h
guptask: The bit-packing format changes for multi-tier compressed pointer. The 32nd bit is reserved for tid only for multi-tiered compressed pointer. When config is single-tiered, compressed pointer bit packing used the original design - 32nd bit is not reserved for tid.
tid ==0 doesn't mean cachelib is not multi-tiered.

cachelib/allocator/memory/CompressedPtr.h
guptask: you're right. I pushed up the correct version now.

cachelib/allocator/memory/CompressedPtr.h
guptask: yes, It will be referred to in subsequent upstream PRs. For complete picture, please refer to https://github.com/intel/CacheLib/pull/56

cachelib/allocator/memory/SlabAllocator.h
guptask: Please refer to this commit https://github.com/intel/CacheLib/pull/56/commits/2704ac894d2a4fcf7be29d38784abf2e0fbc6396#diff-a6542b6dbf2cfb5e03e82205ee960757ab2b50de7bc25085089a3cffba40ae87 which shows how tier Id is passed to the alloctor.compress() methods.

This change is due to be sent for upstream review soon.

therealgymmy:
@guptask: Hey I didn't expect this but I have observed about 1.5% CPU regression. (I did A/B and then B/A in our internal test setups to verify the difference exists).

Have you tried any cachebench runs on your end to see if throughput is affected with this change?

guptask:
> @guptask: Hey I didn't expect this but I have observed about 1.5% CPU regression. (I did A/B and then B/A in our internal test setups to verify the difference exists).
> 
> Have you tried any cachebench runs on your end to see if throughput is affected with this change?

I ran several sets of experiment with compressed_ptr changes (rebased to latest main) and the main branch:
1. graph cache follower fbobj with opRatePerSec set to 1m.
2. CDN workload with opRatePerSec set to 500k.
3. benchmark-test-PtrCompressionBench  with -bm_min_iters=10000

In all experiments, I'm not seeing any regression in stats and CPU usage.

Can you please provide me with more details on how to reproduce this regression ?

guptask:
> > @guptask: Hey I didn't expect this but I have observed about 1.5% CPU regression. (I did A/B and then B/A in our internal test setups to verify the difference exists).
> > Have you tried any cachebench runs on your end to see if throughput is affected with this change?
> 
> @therealgymmy I ran several sets of experiment with compressed_ptr changes (rebased to latest main) and the main branch:
> 
> 1. graph cache follower fbobj with opRatePerSec set to 1m.
> 2. CDN workload with opRatePerSec set to 500k.
> 3. benchmark-test-PtrCompressionBench  with -bm_min_iters=10000
> 
> In all experiments, I'm not seeing any regression in stats and CPU usage.
> 
> Can you please provide me with more details on how to reproduce this regression ?

HI @therealgymmy is there any further update on this regression issue from your end ? I didn't observe any performance regression after rebasing the upstream branch.

therealgymmy:
@guptask: hey i haven't had a chance to re-run the A/B test. Will kick off a run today.

therealgymmy:
Looks like I made a mistake of not building A/B packages close together. Last time my "control" A-side was the production package whereas "test" B-side has many more diffs including this one. This time, A and B only differ in this diff, and the cpu difference is consistent during A/B and after I flipped it to B/A.

Screenshot 2023-04-05 at 11 24 15 AM

therealgymmy (approved):

PR #194, opened 2023-02-18
fix compilation two issues

1a1a11a:
1. remove cmake from Ubuntu apt installation because the cmake installed from apt (both Ubuntu18 and Ubuntu20) is too old to be used, and the new cmake from apt shadows existing cmake binary. 
2. current compilation cannot use multiple cores even though `-j` is specified due to a bug in using `nproc`.

### DIFF
--- a/contrib/build-package.sh
+++ b/contrib/build-package.sh
@@ -259,7 +259,7 @@ test "$debug_build" \
 MAKE_PARAMS=
 test "$verbose" && MAKE_PARAMS="$MAKE_PARAMS VERBOSE=YES"
 
-JOBS=$(nproc --ignore 1)
+JOBS=$(nproc --all --ignore 1)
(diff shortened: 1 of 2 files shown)


therealgymmy:
if we remove cmake from dependency, do we expect users to have installed cmake from a newer source?

wonglkd:
FYI, re cmake being too old: I'm not sure if your cmake issue was the same as mine, but I did face a compilation issue on Ubuntu20 due to zstd (https://github.com/facebook/zstd/issues/3500) using a flag only present in cmake >= 3.18. It seems like they have resolved this issue upstream and the fix should be out in the next release.

(For myself, I resolved my issue by going back to Ubuntu18.)

1a1a11a:
Oh, this is helpful! In that case, then we can keep cmake

1a1a11a:
Oh, this is helpful! In that case, then we can keep cmake.

1a1a11a:
updated

PR #195, opened 2023-02-20
NUMA bindigs support for private memory

vinser52:
This PR adds NUMA binding support for private memory. I tried to align implementation with shared memory and introduced PrivateMemoryManager. And now Slab allocator does not own private memory and probably we have to clean up Slab Allocator as well.

### DIFF
--- a/cachelib/allocator/CMakeLists.txt
+++ b/cachelib/allocator/CMakeLists.txt
@@ -54,6 +54,7 @@ add_library (cachelib_allocator
     PoolOptimizeStrategy.cpp
     PoolRebalancer.cpp
     PoolResizer.cpp
+    PrivateMemoryManager.cpp
     RebalanceStrategy.cpp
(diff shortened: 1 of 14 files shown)


therealgymmy:
@vinser52: apologies for the late review.

Instead of supporting this, I think we should deprecate private memory altogether. Today when a cachelib user has configued memory-monitor (where we may madvise away pages to make more memory available for the rest of the system), we will always use SHM.

This means we will just support Shm for multi-tier memory, and internally we can start moving everyone off private memory mode.

https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheAllocator-inl.h#L53-L54

```
CacheAllocator<CacheTrait>::CacheAllocator(
    typename CacheAllocator<CacheTrait>::InitMemType type, Config config)
    : isOnShm_{type != InitMemType::kNone ? true
                                          : config.memMonitoringEnabled()},
```

PR #161, opened 2022-09-13
NUMA bindings support for Shm Segments

vinser52:
This set of changes allows binding particular memory segments to the specified NUMA node(s).

Issue #135, opened 2022-04-28
CI failures


igchor:
Hi!

I noticed that the CI on main branch and on PRs is failing. Would you be open to integrating CI from our fork (pmem/CacheLib)?

Right now we are building CacheLib and running (almost) all tests on each PR/push on GHA. To speed up things, we are using docker images with pre-installed dependencies. You can see our workflow here: https://github.com/pmem/CacheLib/blob/main/.github/workflows/build-cachelib-docker.yml In our system, we can trigger docker rebuild and push it to the registry once we need to update the dependencies.

agordon:
Thanks, @igchor , I'm investigating this.

PR #198, opened 2023-02-24
Run Github Actions on pull requests

wonglkd:
Run GitHub action builds on every pull request, in addition to currently daily scheduled runs.

Benefit: avoid accidentally breaking other OS builds. This would have caught #197.
This triggers whenever PRs are opened or when commits are added to a PR. 

Frequency of PRs (total of 76 PRs over last 18 months since CacheLib was open-sourced) is much less than frequency  of daily scheduled builds, so this shouldn't add too many builds overall.

### DIFF
--- a/.github/workflows/build-cachelib-centos-8-1.yml
+++ b/.github/workflows/build-cachelib-centos-8-1.yml
@@ -14,6 +14,7 @@
 name: build-cachelib-centos-8-1
 on:
 #  push:
+  pull_request:
   schedule:
      - cron:  '0 11 * * 1,3,5'
 jobs:
--- a/.github/workflows/build-cachelib-centos-8-5.yml
+++ b/.github/workflows/build-cachelib-centos-8-5.yml
@@ -14,6 +14,7 @@
 name: build-cachelib-centos-8.5
 on:
 #   push:
+  pull_request:
   schedule:
      - cron:  '0 9 * * *'
 jobs:
--- a/.github/workflows/build-cachelib-debian-10.yml
+++ b/.github/workflows/build-cachelib-debian-10.yml
@@ -14,6 +14,7 @@
 name: build-cachelib-debian-10
 on:
 #  push:
+  pull_request:
   schedule:
      - cron:  '0 13 * * *'
 jobs:
--- a/.github/workflows/build-cachelib-fedora-36.yml
+++ b/.github/workflows/build-cachelib-fedora-36.yml
@@ -14,6 +14,7 @@
 name: build-cachelib-fedora-36
 on:
 #  push:
+  pull_request:
   schedule:
      - cron:  '0 19 * * *'
 jobs:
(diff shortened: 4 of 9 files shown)


wonglkd:
(Hope you don't mind this PR! Feel free to treat it as a suggestion that you can turn down if you think it results in too many builds. I figured that the frequency of PRs is much less than the frequency of the scheduled builds that already happen, so in my mind the trade-off was worth it.)

jiayuebao:
@wonglkd: Do you mind explaining more why daily scheduled build would cause #197 ? And I just saw you suggested landing #198 before #197, any concern if #197 was landed before #198 ?

wonglkd:
Triggering builds on PR would have avoided the regression fixed by 197, because the build failures would have shown up before #161 was merged into master, and it would have been fixed in PR 161 itself.

No concern, but 198 would have helped to verify if 197 fixed those builds.

therealgymmy:
@wonglkd: thank you for sending this change. Yeah I have been reviewing our github build issues and came to the same conclusion. PR needs to trigger builds. (In fact they should run our tests as well but I haven't figured out a way yet)

jiayuebao:
@wonglkd: From the recent builds https://github.com/facebook/CacheLib/actions: I see both "rockylinux" and "fedora" have passed successfully, so I think the fix of #197 works :) .

For this PR, I do see quite a few build failures - some are caused by missing numa deps while others are not, could you double check all these build failures?

wonglkd:
@jiayuebao I checked the build failures - they're the same as the scheduled ones. There are at least two separate issues (XDCHECK on CentOS[1], and a Docker issue on Debian[2]), and won't be solved by this PR. I did figure out possible solutions to them but think they will be better addressed by their own PRs - I can open an issue to track all the build failures.

Btw, I was thinking that if this is too many builds to trigger, we could still get decent coverage by triggering one of each type (e.g., only CentOS 8.5, only RockyLinux 9.0, etc), but would defer to you re the exact frequency, since I don't have visibility into what build minute budget you have.

I'm less clear on the rpm-build issues and it may involve another repo[3] too  - but I think it makes sense to fix the builds first, and then figure that out after.

[1] I'm running tests on a possible fix in my fork to see if it works - couldn't figure out where in the history it broke too, since it overlapped with the fmt break
[2] https://github.com/actions/runner-images/issues/6775
[3] https://pagure.io/meta/cachelib/blob/packit/f/cachelib.spec

wonglkd:
@therealgymmy sg! Tests would be great. I saw in #135 that pmem got CacheLib CI working with tests, that could be one direction to look at.

jiayuebao:
> they're the same as the scheduled ones. There are at least two separate issues (XDCHECK on CentOS[1], and a Docker issue on Debian[2]), and won't be solved by this PR.

@wonglkd Got it! Thank you for confirming. Yeah no need to include the fix in this PR. 

For the rpm-build issues, our internal team is also aware of that and will work on the fix these days.

wonglkd:
@therealgymmy This is the key script pmem/CacheLib uses to run tests:
https://github.com/pmem/CacheLib/blob/main/run_tests.sh

Could add something like this as a separate step after the build.

Their use of Docker is probably helpful to reduce build time but not essential.

PR #200, opened 2023-02-28
Fix Debian GitHub build & zstd CMake error

wonglkd:
1. Workaround for Debian Docker image bug that is breaking Debian build on GitHub (Explicitly mark Git repo as safe).
2. Pin zstd to a commit that resolves problems with older CMakes (note: affects all OSes, not just Debian)

Context for 1: In latest Debian Docker image , there is a regression that affects the checkout action.

From https://github.com/actions/checkout/issues/1169:
> - Checkout runs, and runs /usr/bin/git config --global --add safe.directory <path>
> - The global .gitconfig does not exist
> - Any calls to git remain unsafe/dubious

The suggested workaround was to use --system instead of --global.

Test Plan: See if GitHub Action Debian build is fixed.

### DIFF
--- a/.github/workflows/build-cachelib-debian-10.yml
+++ b/.github/workflows/build-cachelib-debian-10.yml
@@ -52,6 +52,9 @@ jobs:
           g++ - || true
       - name: "checkout sources"
         uses: actions/checkout@v2
+      - name: "Add Git safe directory"
+        # Workaround for Docker image bug (GitHub issue #199).
+        run: git config --system --add safe.directory /__w/CacheLib/CacheLib
       - name: "Install Prerequisites"
         run: ./contrib/build.sh -S -B
       - name: "Test: update-submodules"
--- a/.github/workflows/build-cachelib-debian-10.yml
+++ b/.github/workflows/build-cachelib-debian-10.yml
@@ -54,7 +54,7 @@ jobs:
         uses: actions/checkout@v2
       - name: "Add Git safe directory"
         # Workaround for Docker image bug (GitHub issue #199).
-        run: git config --system --add safe.directory /__w/CacheLib/CacheLib
+        run: git config --system --add safe.directory $GITHUB_WORKSPACE
       - name: "Install Prerequisites"
         run: ./contrib/build.sh -S -B
       - name: "Test: update-submodules"
--- a/contrib/build-package.sh
+++ b/contrib/build-package.sh
@@ -102,13 +102,12 @@ test "$#" -eq 0 \
     && die "missing dependancy name to build. See -h for help"
 
 ######################################
-## Check which dependecy was requested
+## Check which dependency was requested
 ######################################
 
 external_git_clone=
 external_git_branch=
 external_git_tag=
-external_git_commit=
 update_submodules=
 cmake_custom_params=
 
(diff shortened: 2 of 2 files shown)


wonglkd:
It fixed the issue above, but the Debian build now fails on the cmake/zstd issue. zstd was released 3 weeks ago and then there was a year before that, so we might not see the fix (on dev branch) merged into release any time soon. We could do what is done by [this repo](https://www.github.com/xkszltl/Roaster/commit/ea0074ca0020dedffdd5a57866c3c1f3580c313f), i.e., cherrypicking the commit:

    git fetch origin dev
    git cherry-pick 8420502e

wonglkd:
Decided to pin zstd to the commit that has the cmake fix. Pinning it might be a good idea anyway for stability.

wonglkd:
Debian uses gcc-8.3 and thus now fails on the issue fixed by #201.

contrib/build-package.sh
```diff
@@ -175,7 +176,7 @@ case "$1" in
     REPODIR=cachelib/external/$NAME
     SRCDIR=$REPODIR/build/cmake
     external_git_clone=yes
-    external_git_branch=release
+    external_git_commit=8420502e
```
jaesoo-fb, line 179: Can we simply go with `external_git_tag`? 

This is temporary fix, so I don't think we need to add any complexity. The checkout commands for tag/commit should be identical.

wonglkd: Ah, nice, that's a better solution. Will do, thanks!

wonglkd:
Updated as suggested. Thanks! Also, I got a successful build on my fork after I cherrypicked #201: https://github.com/wonglkd/CacheLib-1/actions/runs/4298872004/jobs/7493484190

contrib/build-package.sh
```diff
@@ -308,6 +311,7 @@ if test "$source" ; then
 
 
     # switch to specific branch/tag if needed
+    # external_git_tag can also be used for commits
```
jaesoo-fb, line 314: Could you move this comment to L180 or L110 where it is more relevant?

.github/workflows/build-cachelib-debian-10.yml
```diff
@@ -52,6 +52,9 @@ jobs:
           g++ - || true
       - name: "checkout sources"
         uses: actions/checkout@v2
+      - name: "Add Git safe directory"
+        # Workaround for Docker image bug (GitHub issue #199).
+        run: git config --system --add safe.directory /__w/CacheLib/CacheLib
```
wonglkd, line 57: Note that this will fail for forks if the fork's name is not CacheLib. Can be resolved by using `/__w/${{github.event.repository.name}}/${{github.event.repository.name}}`, but I worried this was overthinking it especially for a temporary workaround + [some sources](https://stackoverflow.com/questions/62803531/repository-name-as-a-github-action-environment-variable#comment120510045_62805013) say event is unavailable for scheduled workflow pushes.

jaesoo-fb: Oh, not familiar with this, but can we use the [github variable](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables) `$GITHUB_WORKSPACE`? Not sure what is difference against [github context](https://docs.github.com/en/actions/learn-github-actions/contexts) `${{ github.workspace }}`

wonglkd: Ooh, nice find, that's better. I've changed it. Sounds like it's the same, from the page you linked to:

> Most of the default variables have a corresponding, and similarly named, context property

jaesoo-fb (approved):
Thanks for fixing this.

jaesoo-fb:
Somehow I cannot import and merge this. @wonglkd Just in case, could you rebase?

wonglkd:
@jaesoo-fb rebased!

PR #201, opened 2023-02-28
Fix XDCHECK regression in lambda for gcc-8.x

wonglkd:
XDCHECK (or XCHECK) fails here for gcc-8.x in a lambda, so we move it outside. This occurs in CacheBench's AsyncCacheStressor.h.

```
/__w/CacheLib/CacheLib/cachelib/../cachelib/cachebench/runner/AsyncCacheStressor.h:306:7:
internal compiler error: in cp_build_addr_expr_1, at cp/typeck.c:5965
```

This line compiled fine with 8.5 before Jan 2023 so I suspect a regression in folly or other external library. It still compiles fine with gcc-7.5, 9.4, and 11.3.1. Possibly related commits: https://github.com/facebook/folly/commit/1aafad45f316896a7504396f421dacd6c10d7d5f and https://github.com/facebook/folly/commit/e6d09f66b9fc473bc108361d4c8dce8f29f7bcaf

Line of gcc-8.5 that it fails on: https://github.com/gcc-mirror/gcc/blob/releases/gcc-8.5.0/gcc/cp/typeck.c#L5965

The version that isn't in a lambda compiles just fine: https://github.com/facebook/CacheLib/blob/df5b9f6ef35c55e432b6713c52397a03dd19c34c/cachelib/cachebench/runner/CacheStressor.h#L399

Test Plan: GitHub actions. Built fine on my fork with CentOS 8.5/gcc-8.5. This issue currently causes 3 builds that use gcc-8.5 to fail (2 CentOS and RockyLinux-8.6) and 1 build using gcc-8.3 (Debian).

### DIFF
--- a/cachelib/cachebench/runner/AsyncCacheStressor.h
+++ b/cachelib/cachebench/runner/AsyncCacheStressor.h
@@ -23,6 +23,7 @@
 #include <folly/io/async/EventBaseThread.h>
 
 #include <atomic>
+#include <cassert>
 #include <cstddef>
 #include <iostream>
 #include <memory>
@@ -303,7 +304,14 @@ class AsyncCacheStressor : public Stressor {
       } else {
         wHdl = std::move(hdl).toWriteHandle();
       }
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ == 8
+      assert(req->sizeBegin + 1 != req->sizeEnd);
+#else
+      // gcc-8.5 crashes with an internal compiler error at typeck.c:5965
+      // here. Suspected regression in folly as this line compiled fine
+      // with 8.5 previously.
       XDCHECK(req->sizeBegin + 1 != req->sizeEnd);
+#endif
       bool chainSuccessful = false;
       for (auto j = req->sizeBegin + 1; j != req->sizeEnd; j++) {
         ++stats.addChained;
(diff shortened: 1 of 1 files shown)


wonglkd:
All builds now pass, except for Debian, which should be fixed by #200. (Also excluding the rpm-builds.)

cachelib/cachebench/runner/AsyncCacheStressor.h
```diff
@@ -303,7 +304,14 @@ class AsyncCacheStressor : public Stressor {
...
         wHdl = std::move(hdl).toWriteHandle();
       }
+#if defined(__GNUC__) && !defined(__clang__) && __GNUC__ == 8
+      assert(req->sizeBegin + 1 != req->sizeEnd);
+#else
+      // gcc-8.5 crashes with an internal compiler error at typeck.c:5965
+      // here. Suspected regression in folly as this line compiled fine
+      // with 8.5 previously.
       XDCHECK(req->sizeBegin + 1 != req->sizeEnd);
+#endif
```
jaesoo-fb, line 314: It is unfortunate that there is some regression for XDCHECK, but this looks like an overkill. Can you simply move this right before the lambda?

jaesoo-fb (changes requested):

cachelib/cachebench/runner/AsyncCacheStressor.h
wonglkd: Good catch. Done, thanks!

jaesoo-fb (approved):

PR #212, opened 2023-03-13
Expose memory usage statistics for allocation classes

igchor:
And add option to enable printing (for each AC):
- allocSize
- allocated memory size
- memory usage fraction

Memory usage fraction stat is similar to existing 'poolUsageFraction` but per Allocation Class.

Those statistics are used primarily by Background Workers in the multi-tier implementation. They also help understand allocation dynamics.

### DIFF
--- a/cachelib/cachebench/cache/Cache-inl.h
+++ b/cachelib/cachebench/cache/Cache-inl.h
@@ -641,10 +641,20 @@ Stats Cache<Allocator>::getStats() const {
     aggregate += poolStats;
   }
 
+  std::map<PoolId, std::map<ClassId, ACStats>> allocationClassStats{};
+
+  for (size_t pid = 0; pid < pools_.size(); pid++) {
+    auto cids = cache_->getPoolStats(static_cast<PoolId>(pid)).getClassIds();
+    for (auto [cid, stats] :
+         cache_->getPoolStats(static_cast<PoolId>(pid)).mpStats.acStats)
+      allocationClassStats[pid][cid] = stats;
+  }
+
   const auto cacheStats = cache_->getGlobalCacheStats();
   const auto rebalanceStats = cache_->getSlabReleaseStats();
   const auto navyStats = cache_->getNvmCacheStatsMap().toMap();
 
+  ret.allocationClassStats = allocationClassStats;
   ret.numEvictions = aggregate.numEvictions();
   ret.numItems = aggregate.numItems();
   ret.evictAttempts = cacheStats.evictionAttempts;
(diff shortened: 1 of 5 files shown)


cachelib/cachebench/cache/Cache-inl.h
```diff
@@ -620,10 +620,19 @@ Stats Cache<Allocator>::getStats() const {
     aggregate += poolStats;
   }
 
+  std::map<PoolId, std::map<ClassId, ACStats>> allocationClassStats{};
+
+  for (size_t pid = 0; pid < pools_.size(); pid++) {
+    auto cids = cache_->getPoolStats(static_cast<PoolId>(pid)).getClassIds();
+    for (auto cid : cids)
+      allocationClassStats[pid][cid] = cache_->getACStats(pid, cid);
```
haowu14, line 628: I think what you want here is at 
cache_->getPoolStats(poolId).mpStats.acStats is unordered_map<ClassId, ACStats> for the pool.

So you don't have to add a new API into CacheAllocator.h (or Cache)

haowu14 (changes requested):

haowu14:
@igchor  Are you still looking to merge this?

cachelib/cachebench/cache/Cache-inl.h
igchor: Yes, you right, we don't need this extra API. Done.

igchor:
> @igchor Are you still looking to merge this?

Yes, I've rebased the PR and replied to the comment above.

cachelib/allocator/memory/MemoryAllocatorStats.h
```diff
@@ -54,6 +54,17 @@ struct ACStats {
...
+  constexpr double usageFraction() const noexcept {
+    if (usedSlabs == 0)
+      return 0.0;
+
+    return activeAllocs / (usedSlabs * allocsPerSlab);
+  }
+
+  constexpr size_t totalAllocatedSize() const noexcept {
+    return activeAllocs * allocSize;
+  }
```
haowu14, line 67: Are these going to be used inside cachelib by background workers? Otherwise can you move them into cachebench?

haowu14 (changes requested):
Otherwise lgtm.

cachelib/allocator/memory/MemoryAllocatorStats.h
igchor: We wanted to use the in BG workers initially but most likely we will use another mechanism with lower overhead. Done.

PR #213, opened 2023-03-13
Initial framework for background item eviction/promotion

igchor:
Implements stubs for periodic workers for item eviction/promotion. The actual implementation of eviction and promotion is not part of this PR.

The idea behind introducing those workers is the following:
- BG eviction: to keep certain amount of free memory and decrease allocate latency
- BG promotion: to move hot items to memory and decrease read latency

Each BG worker can be customized with an eviction/promotion strategy (not part of this PR). One such strategy is implemented here: https://github.com/intel/CacheLib/blob/develop/cachelib/allocator/FreeThresholdStrategy.cpp

BG promotion is only helpful for multi-tier setup (unless we would have some way of predicting which NVM items are going to be accessed), but BG eviction can be useful for single-tier configuration as well to decrease allocate latency.

### DIFF
--- a/cachelib/allocator/BackgroundMover-inl.h
+++ b/cachelib/allocator/BackgroundMover-inl.h
@@ -1,0 +1,116 @@
+/*
+ * Copyright (c) Intel and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+namespace facebook {
+namespace cachelib {
+
+template <typename CacheT>
+BackgroundMover<CacheT>::BackgroundMover(
+    Cache& cache,
+    std::shared_ptr<BackgroundMoverStrategy> strategy,
+    MoverDir direction)
+    : cache_(cache), strategy_(strategy), direction_(direction) {
+  if (direction_ == MoverDir::Evict) {
+    moverFunc = BackgroundMoverAPIWrapper<CacheT>::traverseAndEvictItems;
+
+  } else if (direction_ == MoverDir::Promote) {
+    moverFunc = BackgroundMoverAPIWrapper<CacheT>::traverseAndPromoteItems;
+  }
+}
+
+template <typename CacheT>
+BackgroundMover<CacheT>::~BackgroundMover() {
+  stop(std::chrono::seconds(0));
+}
+
+template <typename CacheT>
+void BackgroundMover<CacheT>::work() {
+  try {
+    checkAndRun();
+  } catch (const std::exception& ex) {
+    XLOGF(ERR, "BackgroundMover interrupted due to exception: {}", ex.what());
+  }
+}
+
+template <typename CacheT>
+void BackgroundMover<CacheT>::setAssignedMemory(
+    std::vector<MemoryDescriptorType>&& assignedMemory) {
+  XLOG(INFO, "Class assigned to background worker:");
+  for (auto [pid, cid] : assignedMemory) {
+    XLOGF(INFO, "Pid: {}, Cid: {}", pid, cid);
+  }
+
+  mutex_.lock_combine([this, &assignedMemory] {
+    this->assignedMemory_ = std::move(assignedMemory);
+  });
+}
+
+// Look for classes that exceed the target memory capacity
+// and return those for eviction
+template <typename CacheT>
+void BackgroundMover<CacheT>::checkAndRun() {
+  auto assignedMemory = mutex_.lock_combine([this] { return assignedMemory_; });
+
+  unsigned int moves = 0;
+  auto batches = strategy_->calculateBatchSizes(cache_, assignedMemory);
+
+  for (size_t i = 0; i < batches.size(); i++) {
+    const auto [pid, cid] = assignedMemory[i];
+    const auto batch = batches[i];
+
+    if (batch == 0) {
+      continue;
+    }
+
+    // try moving BATCH items from the class in order to reach free target
+    auto moved = moverFunc(cache_, pid, cid, batch);
+    moves += moved;
+    movesPerClass_[pid][cid] += moved;
+    totalBytesMoved_.add(moved * cache_.getPool(pid).getAllocSizes()[cid]);
+  }
+
+  numTraversals_.inc();
+  numMovedItems_.add(moves);
+}
+
+template <typename CacheT>
+BackgroundMoverStats BackgroundMover<CacheT>::getStats() const noexcept {
+  BackgroundMoverStats stats;
+  stats.numMovedItems = numMovedItems_.get();
+  stats.runCount = numTraversals_.get();
+  stats.totalBytesMoved = totalBytesMoved_.get();
+
+  return stats;
+}
+
+template <typename CacheT>
+std::map<PoolId, std::map<ClassId, uint64_t>>
+BackgroundMover<CacheT>::getClassStats() const noexcept {
+  return movesPerClass_;
+}
+
+template <typename CacheT>
+size_t BackgroundMover<CacheT>::workerId(PoolId pid,
+                                         ClassId cid,
+                                         size_t numWorkers) {
+  XDCHECK(numWorkers);
+
+  // TODO: came up with some better sharding (use hashing?)
+  return (pid + cid) % numWorkers;
+}
+
+} // namespace cachelib
+} // namespace facebook
--- a/cachelib/allocator/BackgroundMover.h
+++ b/cachelib/allocator/BackgroundMover.h
@@ -1,0 +1,98 @@
+/*
+ * Copyright (c) Intel and its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "cachelib/allocator/BackgroundMoverStrategy.h"
+#include "cachelib/allocator/CacheStats.h"
+#include "cachelib/common/AtomicCounter.h"
+#include "cachelib/common/PeriodicWorker.h"
+
+namespace facebook {
+namespace cachelib {
+
+// wrapper that exposes the private APIs of CacheType that are specifically
+// needed for the cache api
+template <typename C>
+struct BackgroundMoverAPIWrapper {
+  static size_t traverseAndEvictItems(C& cache,
+                                      unsigned int pid,
+                                      unsigned int cid,
+                                      size_t batch) {
+    return cache.traverseAndEvictItems(pid, cid, batch);
+  }
+
+  static size_t traverseAndPromoteItems(C& cache,
+                                        unsigned int pid,
+                                        unsigned int cid,
+                                        size_t batch) {
+    return cache.traverseAndPromoteItems(pid, cid, batch);
+  }
+};
+
+enum class MoverDir { Evict = 0, Promote };
+
+// Periodic worker that evicts items from tiers in batches
+// The primary aim is to reduce insertion times for new items in the
+// cache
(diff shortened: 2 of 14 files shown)


haowu14:
@igchor  We just upstreamed the shorten critical section PR. Can you rebase this to trunk?

igchor:
> @igchor We just upstreamed the shorten critical section PR. Can you rebase this to trunk?

Sure. Done.

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -3521,6 +3575,68 @@ bool CacheAllocator<CacheTrait>::startNewReaper(
   return true;
 }
 
+template <typename CacheTrait>
+auto CacheAllocator<CacheTrait>::getAssignedMemoryToBgWorker(
+    size_t evictorId, size_t numWorkers) {
```
haowu14, line 3580:
```suggestion
    size_t workerId, size_t numWorkers) {
```

cachelib/allocator/CacheAllocatorConfig.h
```diff
@@ -454,6 +478,18 @@ class CacheAllocatorConfig {
   // make any progress for the below threshold
   std::chrono::milliseconds slabReleaseStuckThreshold{std::chrono::seconds(60)};
 
+  // rebalance to avoid alloc fialures.
```
haowu14, line 481: Please update this doc

cachelib/allocator/BackgroundMover-inl.h
```diff
@@ -0,0 +1,112 @@
...
+  auto batches = strategy_->calculateBatchSizes(cache_, assignedMemory);
+
+  for (size_t i = 0; i < batches.size(); i++) {
+    const auto [pid, cid] = assignedMemory[i];
+    const auto batch = batches[i];
+
+    classes.insert(cid);
+    const auto& mpStats = cache_.getPool(pid).getStats();
+
+    if (!batch) {
```
haowu14, line 78: Can you explain a bit what a batch is? Is it how many items the worker would attempt to move?

cachelib/allocator/BackgroundMover-inl.h
```diff
@@ -0,0 +1,112 @@
...
+      continue;
+    }
+
+    // try moving BATCH items from the class in order to reach free target
+    auto moved = moverFunc(cache_, pid, cid, batch);
+    moves += moved;
+    moves_per_class_[pid][cid] += moved;
+    totalBytesMoved.add(moved * mpStats.acStats.at(cid).allocSize);
```
haowu14, line 86: If all you need is the allocation size, you can use cache_.getPool(pid).getAllocSizes()[cid].

cachelib/allocator/BackgroundMover-inl.h
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  numTraversals.inc();
+  numMovedItems.add(moves);
+  totalClasses.add(classes.size());
```
haowu14, line 91: I think in different memory pools, the same class id can mean different size classes. 
So you may want to count (pid, cid), but wouldn't that just be the size of assignedMemory?

cachelib/allocator/CacheAllocator.h
```diff
@@ -1153,6 +1172,44 @@ class CacheAllocator : public CacheBase {
     return stats;
   }
 
+  // returns the background mover stats
+  BackgroundMoverStats getBackgroundMoverStats(MoverDir direction) const {
```
haowu14, line 1176: could these function be defined in Cache.h, and implemented in CacheAllocator.h like the other getStats functions?

haowu14 (changes requested):

cachelib/allocator/BackgroundMover-inl.h
```diff
@@ -0,0 +1,112 @@
...
+  mutex.lock_combine([this, &assignedMemory] {
+    this->assignedMemory_ = std::move(assignedMemory);
+  });
+}
+
+// Look for classes that exceed the target memory capacity
+// and return those for eviction
+template <typename CacheT>
+void BackgroundMover<CacheT>::checkAndRun() {
+  auto assignedMemory = mutex.lock_combine([this] { return assignedMemory_; });
```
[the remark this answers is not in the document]
igchor: Yes, it might change when a new pool is added (see `addPool`).

cachelib/allocator/BackgroundMover-inl.h
```diff
@@ -0,0 +1,112 @@
...
+  unsigned int moves = 0;
+  std::set<ClassId> classes{};
+  auto batches = strategy_->calculateBatchSizes(cache_, assignedMemory);
+
+  for (size_t i = 0; i < batches.size(); i++) {
+    const auto [pid, cid] = assignedMemory[i];
+    const auto batch = batches[i];
+
+    classes.insert(cid);
+    const auto& mpStats = cache_.getPool(pid).getStats();
```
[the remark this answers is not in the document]
igchor: Done.

cachelib/allocator/CacheAllocator-inl.h
igchor: Done.

cachelib/allocator/BackgroundMover-inl.h
igchor: Yes, that's correct and the batch size is calculated by the BackgroundMoverStrategy - I've added an extra comment there.

cachelib/allocator/BackgroundMover-inl.h
igchor: Done.

cachelib/allocator/BackgroundMover-inl.h
igchor: Yes, I think you are right. I removed this stat for now - we have some other ones implemented in our develop branch that we would probably like to upstream as a next step.

cachelib/allocator/CacheAllocator.h
igchor: I think `getRebalancerStats` etc. are actually not implemented in the Cache.h and I just used the same approach.

cachelib/allocator/BackgroundMover.h
```diff
@@ -0,0 +1,96 @@
...
+    return cache.traverseAndPromoteItems(pid, cid, batch);
+  }
+};
+
+enum class MoverDir { Evict = 0, Promote };
+
+// Periodic worker that evicts items from tiers in batches
+// The primary aim is to reduce insertion times for new items in the
+// cache
+template <typename CacheT>
```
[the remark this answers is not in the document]
igchor: I can rename it but I wanted to expose the alias to the Cache type (through `using Cache = CacheT`) if users of this class would need to know the type.

cachelib/allocator/BackgroundMover-inl.h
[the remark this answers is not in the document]
haowu14: I see.
Let's not loop through all workers, then assign each worker. Let's just loop through (pid, cid) and assign all workers together.

And maybe we have addToAssignMemory for addPool?

haowu14 (changes requested):

cachelib/allocator/CacheAllocator.h
igchor: Actually, I think those functions should just be private. They are only used by `getStats` and `getGlobalStats`.

cachelib/allocator/BackgroundMover-inl.h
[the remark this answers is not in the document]
igchor: I'm not sure if I understand the change you're proposing. How would you like to assign all workers together? We cannot do that under a single lock because lock is per-worker and we need that lock to synchronize `setAssignedMemory` with the `checkAndRun` function. 

As for `addToAssignMemory`, adding a new pool may require rebalancing pools/classes between the workers - that's why I always call `setAssignMemory` and iterate over all workers each time.

cachelib/allocator/CacheAllocator.h
haowu14: ok.

cachelib/allocator/BackgroundMover-inl.h
haowu14: nit, let's be explicit.
```suggestion
    if (batch == 0) {
```

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -3542,6 +3596,68 @@ bool CacheAllocator<CacheTrait>::startNewReaper(
   return true;
 }
 
+template <typename CacheTrait>
+auto CacheAllocator<CacheTrait>::getAssignedMemoryToBgWorker(
+    size_t workerId, size_t numWorkers) {
+  std::vector<MemoryDescriptorType> asssignedMemory;
+  auto pools = filterCompactCachePools(allocator_->getPoolIds());
+  for (const auto pid : pools) {
```
haowu14, line 3604: Can we have a function return vector<vector<MemoryType>>. Then loop through this vector to assign to each worker thread, taking the lock one by one?

haowu14 (changes requested):

cachelib/allocator/CacheAllocator-inl.h
igchor: Sure, done.

cachelib/allocator/BackgroundMover.h
```diff
@@ -0,0 +1,95 @@
...
+  BackgroundMover(Cache& cache,
+                  std::shared_ptr<BackgroundMoverStrategy> strategy,
+                  MoverDir direction_);
+
+  ~BackgroundMover() override;
+
+  BackgroundMoverStats getStats() const noexcept;
+  std::map<PoolId, std::map<ClassId, uint64_t>> getClassStats() const noexcept;
+
+  void setAssignedMemory(std::vector<MemoryDescriptorType>&& assignedMemory);
```
therealgymmy, line 68: This is only needed when we add pools right?

Instead of this, can we stop all background workers and recreate them with the new mem assignments?

Typically, applications do not add a new Pool when it is serving user requests. They typically adjust the pools on startup (e.g. right after recovering the cache from previous run).

cachelib/allocator/CacheAllocator.h
```diff
@@ -710,6 +711,16 @@ class CacheAllocator : public CacheBase {
...
 
+  // create memory assignment to bg workers
+  auto createBgWorkerMemoryAssignments(size_t numWorkers);
+
+  // whether bg worker should be woken
+  bool shouldWakeupBgEvictor(PoolId pid, ClassId cid);
+
+  // bg workers responsible for promoting/evicting from particlar pool and
+  // allocation calss
+  size_t backgroundWorkerId(PoolId pid, ClassId cid, size_t numWorkers);
```
therealgymmy, line 722: can this be a static method in BackgroundMover instead?

cachelib/allocator/FreeThresholdStrategy.cpp
```diff
@@ -0,0 +1,39 @@
...
+                                             uint64_t maxEvictionBatch,
+                                             uint64_t minEvictionBatch)
+    : lowEvictionAcWatermark(lowEvictionAcWatermark),
+      highEvictionAcWatermark(highEvictionAcWatermark),
+      maxEvictionBatch(maxEvictionBatch),
+      minEvictionBatch(minEvictionBatch) {}
+
+std::vector<size_t> FreeThresholdStrategy::calculateBatchSizes(
+    const CacheBase& cache, std::vector<MemoryDescriptorType> acVec) {
+  return {};
```
therealgymmy, line 35: Also add a `throw std::runtime_error("Not supported yet!");` so it's clear this needs to be implemented

therealgymmy (commented):
Looks good! A few comments inline

cachelib/allocator/BackgroundMover.h
igchor: > This is only needed when we add pools right?

Yes, that's true.

> Instead of this, can we stop all background workers and recreate them with the new mem assignments?

We could, but then we would need to introduce extra mutex(es) in CacheAllocator class to protect `backgroundEvictor_` and `backgroundPromoter_` variables. We need to access them for example when we call `backgroundEvictor_[...]->wakeUp()`. Woul you prefer to have that extra mutex instead of the one in BackgroundMover?

cachelib/allocator/CacheAllocator.h
igchor: Sure, done.

cachelib/allocator/FreeThresholdStrategy.cpp
igchor: Done.

cachelib/allocator/BackgroundMover.h
therealgymmy: ah you're right. Actually this made me realize there's a bug in existing CacheAllocator where we try to wake up the Rebalancer after an allocation failure (so we can move slabs around sooner). This actually can crash if user is stopping/creating a new rebalancer at the same time. https://github.com/facebook/CacheLib/blob/main/cachelib/allocator/CacheAllocator-inl.h#L368

----

Let's keep this logic as-is.

----

With the logic as-is today, we still have the race where user calls addPool while they start/stop BackgroundMover worker. Can you add some comments just to clarify that it's not expected that user will call addPool and start/stop workers from two different threads?

therealgymmy (approved):

PR #217, opened 2023-03-24
Shorten critical section in findEviction

igchor:
This is the next part of the 'critical section' patch after https://github.com/facebook/CacheLib/pull/183. 

Original PR (some of the changes already upstreamed): https://github.com/facebook/CacheLib/pull/172

This PR contains changes to findEviction only. The remaining part (changes in SlabRelease code) will be provided in the next (final) PR.

### DIFF
--- a/cachelib/allocator/CacheAllocator-inl.h
+++ b/cachelib/allocator/CacheAllocator-inl.h
@@ -1241,6 +1241,19 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
 }
 
 template <typename CacheTrait>
+void CacheAllocator<CacheTrait>::unlinkItemForEviction(Item& it) {
+  XDCHECK(it.isMarkedForEviction());
+  XDCHECK(it.getRefCount() == 0);
+  accessContainer_->remove(it);
+  removeFromMMContainer(it);
+
+  // Since we managed to mark the item for eviction we must be the only
+  // owner of the item.
+  const auto ref = it.unmarkForEviction();
+  XDCHECK(ref == 0u);
+}
+
+template <typename CacheTrait>
 typename CacheAllocator<CacheTrait>::Item*
 CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
   auto& mmContainer = getMMContainer(pid, cid);
@@ -1248,76 +1261,102 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
   // Keep searching for a candidate until we were able to evict it
   // or until the search limit has been exhausted
   unsigned int searchTries = 0;
-  auto itr = mmContainer.getEvictionIterator();
-  while ((config_.evictionSearchTries == 0 ||
-          config_.evictionSearchTries > searchTries) &&
-         itr) {
-    ++searchTries;
-    (*stats_.evictionAttempts)[pid][cid].inc();
+  while (config_.evictionSearchTries == 0 ||
+         config_.evictionSearchTries > searchTries) {
+    Item* toRecycle = nullptr;
+    Item* candidate = nullptr;
+    typename NvmCacheT::PutToken token;
 
-    Item* toRecycle = itr.get();
+    mmContainer.withEvictionIterator([this, pid, cid, &candidate, &toRecycle,
+                                      &searchTries, &mmContainer,
+                                      &token](auto&& itr) {
+      if (!itr) {
+        ++searchTries;
+        (*stats_.evictionAttempts)[pid][cid].inc();
+        return;
+      }
 
-    Item* candidate =
-        toRecycle->isChainedItem()
-            ? &toRecycle->asChainedItem().getParentItem(compressor_)
-            : toRecycle;
+      while ((config_.evictionSearchTries == 0 ||
+              config_.evictionSearchTries > searchTries) &&
+             itr) {
+        ++searchTries;
+        (*stats_.evictionAttempts)[pid][cid].inc();
 
-    // make sure no other thead is evicting the item
-    if (candidate->getRefCount() != 0 || !candidate->markMoving()) {
-      ++itr;
+        auto* toRecycle_ = itr.get();
+        auto* candidate_ =
+            toRecycle_->isChainedItem()
+                ? &toRecycle_->asChainedItem().getParentItem(compressor_)
+                : toRecycle_;
+
+        const bool evictToNvmCache = shouldWriteToNvmCache(*candidate_);
+        if (evictToNvmCache)
+          token = nvmCache_->createPutToken(candidate_->getKey());
+
+        if (evictToNvmCache && !token.isValid()) {
+          stats_.evictFailConcurrentFill.inc();
+        } else if (candidate_->markForEviction()) {
+          XDCHECK(candidate_->isMarkedForEviction());
+          // markForEviction to make sure no other thead is evicting the item
+          // nor holding a handle to that item
+          toRecycle = toRecycle_;
+          candidate = candidate_;
+
+          // Check if parent changed for chained items - if yes, we cannot
+          // remove the child from the mmContainer as we will not be evicting
+          // it. We could abort right here, but we need to cleanup in case
+          // unmarkForEviction() returns 0 - so just go through normal path.
+          if (!toRecycle_->isChainedItem() ||
+              &toRecycle->asChainedItem().getParentItem(compressor_) ==
+                  candidate)
+            mmContainer.remove(itr);
+          return;
+        } else {
+          if (candidate_->hasChainedItem()) {
+            stats_.evictFailParentAC.inc();
+          } else {
+            stats_.evictFailAC.inc();
+          }
+        }
+
+        ++itr;
+        XDCHECK(toRecycle == nullptr);
+        XDCHECK(candidate == nullptr);
+      }
+    });
+
+    if (!toRecycle)
       continue;
+
+    XDCHECK(toRecycle);
+    XDCHECK(candidate);
+
+    unlinkItemForEviction(*candidate);
+
+    if (token.isValid() && shouldWriteToNvmCacheExclusive(*candidate)) {
+      nvmCache_->put(*candidate, std::move(token));
     }
 
-    // for chained items, the ownership of the parent can change. We try to
-    // evict what we think as parent and see if the eviction of parent
-    // recycles the child we intend to.
-    bool evictionSuccessful = false;
-    {
-      auto toReleaseHandle =
-          itr->isChainedItem()
-              ? advanceIteratorAndTryEvictChainedItem(itr)
-              : advanceIteratorAndTryEvictRegularItem(mmContainer, itr);
-      evictionSuccessful = toReleaseHandle != nullptr;
-      // destroy toReleaseHandle. The item won't be released to allocator
-      // since we marked for eviction.
-    }
-
-    const auto ref = candidate->unmarkMoving();
-    if (ref == 0u) {
-      // Invalidate iterator since later on we may use this mmContainer
-      // again, which cannot be done unless we drop this iterator
-      itr.destroy();
-
-      // recycle the item. it's safe to do so, even if toReleaseHandle was
-      // NULL. If `ref` == 0 then it means that we are the last holder of
-      // that item.
-      if (candidate->hasChainedItem()) {
-        (*stats_.chainedItemEvictions)[pid][cid].inc();
-      } else {
-        (*stats_.regularItemEvictions)[pid][cid].inc();
-      }
-
-      if (auto eventTracker = getEventTracker()) {
-        eventTracker->record(AllocatorApiEvent::DRAM_EVICT, candidate->getKey(),
-                             AllocatorApiResult::EVICTED, candidate->getSize(),
-                             candidate->getConfiguredTTL().count());
-      }
-
-      // check if by releasing the item we intend to, we actually
-      // recycle the candidate.
-      if (ReleaseRes::kRecycled ==
-          releaseBackToAllocator(*candidate, RemoveContext::kEviction,
-                                 /* isNascent */ false, toRecycle)) {
-        return toRecycle;
-      }
+    // recycle the item. it's safe to do so, even if toReleaseHandle was
+    // NULL. If `ref` == 0 then it means that we are the last holder of
+    // that item.
+    if (candidate->hasChainedItem()) {
+      (*stats_.chainedItemEvictions)[pid][cid].inc();
     } else {
-      XDCHECK(!evictionSuccessful);
+      (*stats_.regularItemEvictions)[pid][cid].inc();
     }
 
-    // If we destroyed the itr to possibly evict and failed, we restart
-    // from the beginning again
-    if (!itr) {
-      itr.resetToBegin();
+    if (auto eventTracker = getEventTracker()) {
+      eventTracker->record(AllocatorApiEvent::DRAM_EVICT, candidate->getKey(),
+                           AllocatorApiResult::EVICTED, candidate->getSize(),
+                           candidate->getConfiguredTTL().count());
+    }
+
+    // check if by releasing the item we intend to, we actually
+    // recycle the candidate.
+    auto ret = releaseBackToAllocator(*candidate, RemoveContext::kEviction,
+                                      /* isNascent */ false, toRecycle);
+    if (ret == ReleaseRes::kRecycled) {
+      return toRecycle;
     }
   }
   return nullptr;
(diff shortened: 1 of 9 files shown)


cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1269,100 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+        }
+
+        ++itr;
+        XDCHECK(toRecycle == nullptr);
+        XDCHECK(candidate == nullptr);
       }
+    });
+
+    if (!toRecycle)
+      continue;
```
haowu14, line 1334:
```suggestion
    if (!toRecycle) {
      continue;
    }
```

haowu14: This happens iff we exhausted the entire linked list and hasn't found a candidate while not exhausting searchTries right?

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1269,100 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+          // markForEviction to make sure no other thead is evicting the item
+          // nor holding a handle to that item
+          toRecycle = toRecycle_;
+          candidate = candidate_;
+
+          // Check if parent changed for chained items - if yes, we cannot
+          // remove the child from the mmContainer as we will not be evicting
+          // it. We could abort right here, but we need to cleanup in case
+          // unmarkForEviction() returns 0 - so just go through normal path.
+          if (!toRecycle_->isChainedItem() ||
```
haowu14, line 1314: Can you elaborate more?
In the old way of eviction, if we are trying to recycle a chained item (toRecycle.isChainedItem()) but the parent has changed, we don't do any specific treatment until in releaseBackToAllocator, we'll go through the chain with parent being `candidate`, and found out that `toRecycle` was never reached by walking the chain. So releaseBackToAllocator fails and we continue iterating the eviction queue.

How does the new way work?

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1269,100 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+
+        auto* toRecycle_ = itr.get();
+        auto* candidate_ =
+            toRecycle_->isChainedItem()
+                ? &toRecycle_->asChainedItem().getParentItem(compressor_)
+                : toRecycle_;
+
+        token = createPutToken(*candidate_);
+
+        if (shouldWriteToNvmCache(*candidate_) && !token.isValid()) {
```
haowu14, line 1301: `shouldWriteToNvmCache` is called once in `createPutToken` and another time in here, can we make it called only once? I think we can just do it as the old way?

haowu14 (changes requested):

cachelib/allocator/CacheAllocator-inl.h
igchor: Since I wrapped token creation in `createPutToken` I no longer know if token is invalid due to `FailConcurrentFill` or due because `shouldWriteToNvmCache` returned false.

Perhaps I should just put `stats_.evictFailConcurrentFill.inc();` inside `createPutToken`?

cachelib/allocator/CacheAllocator-inl.h
igchor: The new way is pretty much the same. This comment states that we don't do any specific treatment as well. We'll just try to evict the item as usual and leverage the existing logic in releaseBackToAllocator.

It might be possible to optimize this case but for this PR I felt like having similar behavior is better.

The only reason for this `if` here is to optimize the happy case: if the item is not a chained item or the parent hasn't changed, we can safely remove the item from MMContainer (we already hold the MMContainer lock so doing it here has low overhead). If the parent has changed we cannot call the `mmContainer.remove()` here as it could lead to deadlocks. Instead, we'll just call `mmContainer.remove()` inside `unlinkItemForEviction` which is slower (need to acquire the lock again) but safe.

cachelib/allocator/CacheAllocator-inl.h
igchor: Yes, that's correct.

cachelib/allocator/CacheAllocator-inl.h
haowu14: Is there a reason you have to create a function for createPutToken?

cachelib/allocator/CacheAllocator-inl.h
haowu14: I understand the optimization now. 
I think as long as the chained item can't change parent after this point then this is safe. And since we marked the parent as forEviction, it can't be moving nor accessed, so we don't have to do the chainedItemLock_ here.

How much does this optimization give us?

haowu14 (changes requested):

cachelib/allocator/CacheAllocator-inl.h
igchor: I wanted to use it in other places as well. But since for now it's only used in findEviction I can remove it. Done.

igchor: Also, I just realized I bumped `evictFailParentAC` and `evictFailAC` even if the failure was due to invalid token - I fixed this as well.

cachelib/allocator/CacheAllocator-inl.h
igchor: Using default `hit_ratio/graph_cache_leader_fbobj/config.json` this optimization improves throughput by ~40% on my machine. This is mostly due to almost 100% eviction Success rate (without the optimization I get ~50% Success rate).

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1261,102 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+        auto* toRecycle_ = itr.get();
+        auto* candidate_ =
+            toRecycle_->isChainedItem()
+                ? &toRecycle_->asChainedItem().getParentItem(compressor_)
+                : toRecycle_;
+
+        const bool evictToNvmCache = shouldWriteToNvmCache(*candidate_);
+        if (evictToNvmCache)
+          token = nvmCache_->createPutToken(candidate_->getKey());
+
```
haowu14, line 1294:
```suggestion
        if (evictToNvmCache) {
          token = nvmCache_->createPutToken(candidate_->getKey());
        }

```

haowu14: And this might have been lost in one of the patches you add after you ran test but here is a bug.
The token can be created on candidate A, then candidate A failed to markForEviction and we move to the next item.
For candidate B, evictToNvmCache is false but we successfully markForEviction. 

We'll exit this loop with candidate B but the token created from candidate A.

cachelib/allocator/CacheAllocator-inl.h
haowu14: Please fix the format suggestion.

haowu14 (changes requested):

cachelib/allocator/CacheAllocator-inl.h
haowu14: I saw a crash in production test with exception "Item is not nvm evicted and nvm clean". And I suspect this is the reason. I'm putting together a hacky fix on my end to resume the test.

haowu14 (changes requested):
@igchor I continued the production test with a hacky change where I do `auto tokoen_ = createPutToken(...)` and only do `token == std::move(token)` iff the item successfully mark for eviction. However, I noticed a regression in allocation latency mostly caused by creating and destroying PutTokens.
I think this is because in this PR, we try to create put token before mark for eviction while before this PR we do refcount==0 and markMoving before trying to create PutToken.
Can we do it the same way as before?

igchor:
> @igchor I continued the production test with a hacky change where I do `auto tokoen_ = createPutToken(...)` and only do `token == std::move(token)` iff the item successfully mark for eviction. However, I noticed a regression in allocation latency mostly caused by creating and destroying PutTokens. I think this is because in this PR, we try to create put token before mark for eviction while before this PR we do refcount==0 and markMoving before trying to create PutToken. Can we do it the same way as before?

We need to create token before marking for eviction because `markForEviction` is the point where the item stops being visible to other threads. I think we could do some optimizations here: for example, check the refcount / moving bit first, and move to the next item if necessary (it's still possible we'll need to destroy the token but it should be less common).

Could you share what is the  `Evict Success` rate and how much regression have you observed?  I've run a few benchmarks where evict success rate was around 80% and didn't see any regression with the fix. I've also included the fix in this PR as a separate commit.

Also, have you tried running with "useCombinedLockForIterators": true ?

haowu14:
@igchor 
I just realized I made a mistake in my hacky fix. I just fixed that and am rebuilding a binary. The mistake I made was checking the wrong token, which would always be invalid and resulted in the advance of iterator. Hopefully this de-signify the fact that now we create token before mark evict.

> "useCombinedLockForIterators": true

Not yet. I'd need to make sure with this being false there's no regression (I hope)?

igchor:
> @igchor I just realized I made a mistake in my hacky fix. I just fixed that and am rebuilding a binary. The mistake I made was checking the wrong token, which would always be invalid and resulted in the advance of iterator. Hopefully this de-signify the fact that now we create token before mark evict.
> 
> > "useCombinedLockForIterators": true
> 
> Not yet. I'd need to make sure with this being false there's no regression (I hope)?

Got it. Right, regarding useCombinedLockForIterators, I would expect there is no regression when it's set to false. When it's set to 'true' I'd expect lower allocate latency and higher throughput at the cost of slightly higher find latency.

haowu14:
In the most recent version there's no significant regression however I don't see improvement with combined locking on either.

Our test is a shadow set tup, meaning that the binary running base (trunk) and test(this PR) will receive exact same get/set at the same time regardless of how quickly that binary can process the request. So we can't look at QPS directly for comparison.
Looking at p95 allocation latency, they look pretty much the same (test is about 1us slower, not sure if it's just accounting error). The eviction success rate is between 55 ~ 70 and they are the same most of the time across the two versions.

I'll run the experiment again on a new experiment since this existing pair of servers ran my buggy code and may not have exactly the same cache content (but I don't think it matters much).

My questions for you:
1. Any other metrics you'd like me to look at to use as a poxy to evaluate this throughput win?
2. Can you remind me of what's the slow down on find path?

igchor:
> In the most recent version there's no significant regression however I don't see improvement with combined locking on either.
> 
> Our test is a shadow set tup, meaning that the binary running base (trunk) and test(this PR) will receive exact same get/set at the same time regardless of how quickly that binary can process the request. So we can't look at QPS directly for comparison. Looking at p95 allocation latency, they look pretty much the same (test is about 1us slower, not sure if it's just accounting error). The eviction success rate is between 55 ~ 70 and they are the same most of the time across the two versions.
> 
> I'll run the experiment again on a new experiment since this existing pair of servers ran my buggy code and may not have exactly the same cache content (but I don't think it matters much).
> 
> My questions for you:
> 
> 1. Any other metrics you'd like me to look at to use as a poxy to evaluate this throughput win?
> 2. Can you remind me of what's the slow down on find path?

1. We actually saw the highest difference on p99 (and higher) latencies. 
I think combined locking helps the most if there is higher contention on the MMContainer lock. If it's not the case for the workload you're running, then there might no difference. However, combined locking would still probably help if our multi-tier version would be run on that workload (the length of the critical section for multi-tier is longer which causes contention to go up).

2. MMContainer lock is used in (at least) 2 places: `findEviction` (called from `allocate`) to protect the search for an eviction candidate and in `markUseful` which is called by all readers (on `find`). Without combined locking, a thread calling `find` would just execute its (short) critical section or wait in case of contention. With combined locking, that thread might end up executing a queued task from `findEviction` as well, and since the `findEviction` critical section length is longer, it might increase `find` latency. My understanding is that with combined locking we are pretty much splitting the work (under MMContainer lock) more equally between `allocate` and `find`

Also, one more thing that we've noticed when running benchmarks. With this patch, we can get similar (or even better) performance (QPS and latency) while using lower AC `locksPower_` (because we no longer lock AC under MMContainer lock).

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1240,6 +1240,19 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
   return true;
 }
 
+template <typename CacheTrait>
+void CacheAllocator<CacheTrait>::unlinkItemForEviction(Item& it) {
+  XDCHECK(it.isMarkedForEviction());
+  XDCHECK(it.getRefCount() == 0);
```
therealgymmy, line 1246: nitpick: XDCHECK_EQ(0, it.getRefcount());

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1240,6 +1240,19 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
...
+  accessContainer_->remove(it);
+  removeFromMMContainer(it);
```
therealgymmy, line 1248: didn't we already remove this item from mm-container in findEviction at L1313? (if it were a parent or just a regular item?)

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1240,6 +1240,19 @@ bool CacheAllocator<CacheTrait>::moveChainedItem(ChainedItem& oldItem,
...
+
+  // Since we managed to mark the item for eviction we must be the only
+  // owner of the item.
+  const auto ref = it.unmarkForEviction();
+  XDCHECK(ref == 0u);
```
therealgymmy, line 1253: same nitpick: XDCHECK_EQ

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1261,106 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
+
+        const bool evictToNvmCache = shouldWriteToNvmCache(*candidate_);
+        auto token_ = evictToNvmCache
+                          ? nvmCache_->createPutToken(candidate_->getKey())
+                          : typename NvmCacheT::PutToken{};
+
+        if (evictToNvmCache && !token_.isValid()) {
+          stats_.evictFailConcurrentFill.inc();
+        } else if (candidate_->markForEviction()) {
+          XDCHECK(candidate_->isMarkedForEviction());
```
therealgymmy, line 1299: Can we refactor this a bit to have fewer nested if-stmt?

e.g. 
```
if (evictToNvmCache && !token_.isValid()) {
  stats_.evictFailConcurrentFill.inc();
  ++itr;
  continue;
}

auto markedForEviction = candidate_->markForEviction();
if (!markedForEviction) {
  ++itr;
  continue;
}

// ...
```

cachelib/allocator/CacheAllocator-inl.h
```diff
@@ -1248,76 +1261,106 @@ CacheAllocator<CacheTrait>::findEviction(PoolId pid, ClassId cid) {
...
-    Item* candidate =
-        toRecycle->isChainedItem()
-            ? &toRecycle->asChainedItem().getParentItem(compressor_)
-            : toRecycle;
-
-    // make sure no other thead is evicting the item
-    if (candidate->getRefCount() != 0 || !candidate->markMoving()) {
-      ++itr;
+  while (config_.evictionSearchTries == 0 ||
+         config_.evictionSearchTries > searchTries) {
```
therealgymmy, line 1265: Do we still need this when we're doing a while loop already inside `withEvictionIterator`?

Can we just reset the iterator internally and restart the search under lock?

therealgymmy (commented):
Sorry for the late review. This LGTM overall. A few comments inline.

cachelib/allocator/CacheAllocator-inl.h
igchor: Potentially yes, but it might have an impact on performance. The current implementation behaves pretty much the same as the original implementation. If we got rid of this loop, the MMContainer lock could be held for a longer duration.

Resetting the iterator under the lock might actually be a better solution but I just didn't want to change the behavior too much in this PR - perhaps it would be better to do it in a separate PR and evaluate it?

igchor: Also, I think the outer loop is also need for cases when releaseBackToAllocator ends up not releasing the item we intend to - this is already done outside the lock.

cachelib/allocator/CacheAllocator-inl.h

cachelib/allocator/CacheAllocator-inl.h
igchor: Yes, we might have removed the item but `removeFromMMContainer` checks if the item is still in the MMContainer and only then calls mmcontainer.remove(). We need to keep this call here because If the parent of the item changed, the item will still be in the mmcontainer.

cachelib/allocator/CacheAllocator-inl.h
igchor: Done

cachelib/allocator/CacheAllocator-inl.h
igchor: Sure, done.

haowu14:
Performance tests have passed with autoFDO off. So we are good performance wise.

haowu14:
Crashed in staging test for a service. Validating a fix. We'll land it again once it's fixed.

haowu14:
We upstreamed this again https://github.com/facebook/CacheLib/commit/ae7442ab1832bad98bfaaaeaa377727a80cf56b4

The problem was the lifetime of putToken. It could live after the `candidate` is released and its key becomes empty. The destructor would try to remove this empty key from the InflightPuts map and assert for success. It's fixed by making sure the putToken is destroyed right after attempting to insert into NVM, before trying to release candidate back to allocator.

igchor:
@haowu14 thanks for letting us now! We'll try to apply the fix on our end and validate it as well. We've actually seen a few `inconsistencies` reported by cachebench the last few days on our develop branch but we thought this was due to other changes.

PR #246, opened 2023-06-26
Rolling Average Latency Tracker

guptask:
This is a follow-up on the existing stats PR (https://github.com/facebook/CacheLib/pull/212)
Added a Rolling Latency tracker class that is used to calculate the rolling average allocation latency. This feature is low overhead and useful for scenarios where frequent calculation of histogram-based tail latency adds performance overhead.


haowu14:
I thought we have a histogram-based latency tracker already enabled for the allocation. Is there any reason to add this one?

PR #254, opened 2023-09-12
modified for building Cachebench on CentOS9

shadow-daylight:
When building in CentOS9 had found the path missing (found there was an additional path compared to other build function)


jaesoo-fb:
Thanks for reporting this issue. We are aware of this and let us piggy-back this to the patch fixing the same issue for Centos Stream 9 build.

PR #258, opened 2023-09-21
Fix modification of it in the loop

JasonGuo98:
Update the code to correctly modify the values of AccuTailHitsRegularPool[pid] in the loop. Previously, the code was modifying a copy of the value instead of the original value.


therealgymmy:
thanks for the fix. How did you come across this bug?

PR #277, opened 2023-11-15
(PR2) Adds the support for Flexible Data Placement(FDP) over NVMe into the Cachelib

arungeorge83:
This adds the device layer support for NVMe-FDP semantics and adds the RUH-awareness feature of NVMe-FDP in
the upper layers of Navy. This allows the BlockCache(large items) and BigHash(small items) of Navy to segregate their data streams in physical NAND media by using the FDP placement Identifiers.

With this changes, the Cachelib can reduce the Device Write Amplification (WAF) significantly even in high SSD utilization scenarios("nvmCacheSizeMB" above 50% of the SSD capacity) in most of the cachelib workloads.

This commit introduces a 'placementHandle' concept for data placement, which can be used by both BC and BH of Navy on
device write() calls, especially for FDP placements. The 'placementHandle' have to be allocated from the device.

io_uring_cmd interface(through nvme char device) is used to send FDP directives to Linux kernel, as sending it through the
conventional block interfaces is not suported yet. The user can select the NVMe block device (Namespace/partition) as usual
(Ex: "nvmCachePaths": ["/dev/nvme0n1p1"]), and the cachelib will pick the corresponding NVMe char device internally.

This commit adds a new config 'fdpMode' to enable FDP. The user needs to select the fdpMode along with iOUring I/O Engine options ("navyEnableIoUring": true, "navyQDepth": 1, "fdpMode": true).

This second PR consist of the code which can be integrated to the 'iouring async-io' support in the Navy.

### DIFF
--- a/cachelib/allocator/nvmcache/NavyConfig.h
+++ b/cachelib/allocator/nvmcache/NavyConfig.h
@@ -485,6 +485,8 @@ class NavyConfig {
   bool isBigHashEnabled() const {
     return enginesConfigs_[0].bigHash().getSizePct() > 0;
   }
+  bool isFDPEnabled() const { return enableFDP_; }
+
   std::map<std::string, std::string> serialize() const;
 
   // Getters:
@@ -549,6 +551,8 @@ class NavyConfig {
   // ============ Device settings =============
   // Set the device block size, i.e., minimum unit of IO
   void setBlockSize(uint64_t blockSize) noexcept { blockSize_ = blockSize; }
+  // Set the NVMe FDP Device data placement mode in the Cachelib
+  void setEnableFDP(bool enable) noexcept { enableFDP_ = enable; }
   // Set the parameters for a simple file.
   // @throw std::invalid_argument if RAID files have been already set.
   void setSimpleFile(const std::string& fileName,
@@ -694,6 +698,9 @@ class NavyConfig {
   // Whether to use write size (instead of parcel size) for Navy admission
   // policy.
   bool useEstimatedWriteSize_{false};
+  // Whether Navy support the NVMe FDP data placement(TP4146) directives or not.
+  // Reference: https://nvmexpress.org/nvmeflexible-data-placement-fdp-blog/
+  bool enableFDP_{false};
 };
 } // namespace navy
 } // namespace cachelib
--- a/cachelib/navy/Factory.cpp
+++ b/cachelib/navy/Factory.cpp
@@ -454,6 +454,7 @@ std::unique_ptr<Device> createFileDevice(
     uint32_t maxDeviceWriteSize,
     IoEngine ioEngine,
     uint32_t qDepth,
+    bool isFDPEnabled,
     std::shared_ptr<navy::DeviceEncryptor> encryptor) {
   // File paths are opened in the increasing order of the
   // path string. This ensures that RAID0 stripes aren't
@@ -476,12 +477,14 @@ std::unique_ptr<Device> createFileDevice(
   }
 
   return createDirectIoFileDevice(std::move(fileVec),
+                                  std::move(filePaths),
                                   fdSize,
                                   blockSize,
                                   stripeSize,
                                   maxDeviceWriteSize,
                                   ioEngine,
                                   qDepth,
+                                  isFDPEnabled,
                                   std::move(encryptor));
 }
 
--- a/cachelib/navy/bighash/BigHash.cpp
+++ b/cachelib/navy/bighash/BigHash.cpp
@@ -88,7 +88,8 @@ BigHash::BigHash(Config&& config, ValidConfigTag)
       cacheBaseOffset_{config.cacheBaseOffset},
       numBuckets_{config.numBuckets()},
       bloomFilter_{std::move(config.bloomFilter)},
-      device_{*config.device} {
+      device_{*config.device},
+      placementHandle_{device_.allocatePlacementHandle()} {
   XLOGF(INFO,
         "BigHash created: buckets: {}, bucket size: {}, base offset: {}",
         numBuckets_,
@@ -550,6 +551,7 @@ Buffer BigHash::readBucket(BucketId bid) {
 bool BigHash::writeBucket(BucketId bid, Buffer buffer) {
   auto* bucket = reinterpret_cast<Bucket*>(buffer.data());
   bucket->setChecksum(Bucket::computeChecksum(buffer.view()));
-  return device_.write(getBucketOffset(bid), std::move(buffer));
+  return device_.write(
+      getBucketOffset(bid), std::move(buffer), placementHandle_);
 }
 } // namespace facebook::cachelib::navy
--- a/cachelib/navy/common/Device.cpp
+++ b/cachelib/navy/common/Device.cpp
@@ -31,6 +31,7 @@
 #include <cstring>
 #include <numeric>
 
+#include "cachelib/navy/common/FdpNvme.h"
 #include "cachelib/navy/common/Utils.h"
 
 namespace facebook::cachelib::navy {
@@ -57,13 +58,15 @@ struct IOOp {
                 int fd,
                 uint64_t offset,
                 uint32_t size,
-                void* data)
+                void* data,
+                int handle)
       : parent_(parent),
         idx_(idx),
         fd_(fd),
         offset_(offset),
         size_(size),
-        data_(data) {}
+        data_(data),
+        handle_(handle) {}
 
   std::string toString() const;
 
@@ -78,6 +81,7 @@ struct IOOp {
   const uint64_t offset_ = 0;
   const uint32_t size_ = 0;
   void* const data_;
+  int handle_;
 
   // The number of resubmission on EAGAIN error
   uint8_t resubmitted_ = 0;
@@ -97,7 +101,8 @@ struct IOReq {
                  OpType opType,
                  uint64_t offset,
                  uint32_t size,
-                 void* data);
+                 void* data,
+                 int handle);
 
   const char* getOpName() const {
     switch (opType_) {
@@ -129,6 +134,7 @@ struct IOReq {
   const uint64_t offset_ = 0;
   const uint32_t size_ = 0;
   void* const data_;
+  int handle_;
 
   // Aggregate result of operations
   bool result_ = true;
@@ -167,7 +173,8 @@ class IoContext {
                                      uint32_t stripeSize,
                                      uint64_t offset,
                                      uint32_t size,
-                                     const void* data);
+                                     const void* data,
+                                     int handle);
 
   // Submit a IOOp to the device; should not fail for AsyncIoContext
   virtual bool submitIo(IOOp& op) = 0;
@@ -217,7 +224,8 @@ class AsyncIoContext : public IoContext {
                  size_t id,
                  folly::EventBase* evb,
                  size_t capacity,
-                 bool useIoUring);
+                 bool useIoUring,
+                 std::vector<std::shared_ptr<FdpNvme>> fdpNvmeVec);
 
   ~AsyncIoContext() override = default;
 
@@ -233,6 +241,11 @@ class AsyncIoContext : public IoContext {
 
  private:
   void handleCompletion(folly::Range<folly::AsyncBaseOp**>& completed);
+
+  std::unique_ptr<folly::AsyncBaseOp> prepAsyncIo(IOOp& op);
+
+  // Prepare an Nvme CMD IO through IOUring
+  std::unique_ptr<folly::AsyncBaseOp> prepNvmeIo(IOOp& op);
 
   // The maximum number of retries when IO failed with EBUSY. 5 is arbitrary
   static constexpr size_t kRetryLimit = 5;
@@ -260,6 +273,11 @@ class AsyncIoContext : public IoContext {
   size_t numOutstanding_ = 0;
   size_t numSubmitted_ = 0;
   size_t numCompleted_ = 0;
+
+  // Device info vector for FDP support
+  const std::vector<std::shared_ptr<FdpNvme>> fdpNvmeVec_{};
+  // As of now, only one FDP enabled Device is supported
+  static constexpr uint16_t kDefaultFdpIdx = 0u;
 };
 
 // An FileDevice manages direct I/O to either a single or multiple (RAID0)
@@ -267,9 +285,11 @@ class AsyncIoContext : public IoContext {
 class FileDevice : public Device {
  public:
   FileDevice(std::vector<folly::File>&& fvec,
+             std::vector<std::shared_ptr<FdpNvme>>&& fdpNvmeVec,
              uint64_t size,
              uint32_t blockSize,
              uint32_t stripeSize,
+             uint32_t maxIOSize,
              uint32_t maxDeviceWriteSize,
              IoEngine ioEngine,
              uint32_t qDepthPerContext,
@@ -281,14 +301,19 @@ class FileDevice : public Device {
  private:
   IoContext* getIoContext();
 
-  bool writeImpl(uint64_t, uint32_t, const void*) override;
+  bool writeImpl(uint64_t, uint32_t, const void*, int) override;
 
   bool readImpl(uint64_t, uint32_t, void*) override;
 
   void flushImpl() override;
 
+  int allocatePlacementHandle() override;
+
   // File vector for devices or regular files
   const std::vector<folly::File> fvec_{};
+
+  // Device info vector for FDP support
+  const std::vector<std::shared_ptr<FdpNvme>> fdpNvmeVec_{};
 
   // RAID stripe size when multiple devices are used
   const uint32_t stripeSize_;
@@ -321,7 +346,7 @@ class MemoryDevice final : public Device {
   explicit MemoryDevice(uint64_t size,
                         std::shared_ptr<DeviceEncryptor> encryptor,
                         uint32_t ioAlignSize)
-      : Device{size, std::move(encryptor), ioAlignSize,
+      : Device{size, std::move(encryptor), ioAlignSize, 0 /* max IO size */,
                0 /* max device write size */},
         buffer_{std::make_unique<uint8_t[]>(size)} {}
   MemoryDevice(const MemoryDevice&) = delete;
@@ -331,7 +356,8 @@ class MemoryDevice final : public Device {
  private:
   bool writeImpl(uint64_t offset,
                  uint32_t size,
-                 const void* value) noexcept override {
+                 const void* value,
+                 int /* unused */) noexcept override {
     XDCHECK_LE(offset + size, getSize());
     std::memcpy(buffer_.get() + offset, value, size);
     return true;
@@ -343,6 +369,8 @@ class MemoryDevice final : public Device {
     return true;
   }
 
+  int allocatePlacementHandle() override { return -1; }
+
   void flushImpl() override {
     // Noop
   }
@@ -351,20 +379,20 @@ class MemoryDevice final : public Device {
 };
 } // namespace
 
-bool Device::write(uint64_t offset, BufferView view) {
+bool Device::write(uint64_t offset, BufferView view, int handle) {
   if (encryptor_) {
     auto writeBuffer = makeIOBuffer(view.size());
     writeBuffer.copyFrom(0, view);
-    return write(offset, std::move(writeBuffer));
+    return write(offset, std::move(writeBuffer), handle);
   }
 
   const auto size = view.size();
   XDCHECK_LE(offset + size, size_);
   const uint8_t* data = reinterpret_cast<const uint8_t*>(view.data());
-  return writeInternal(offset, data, size);
+  return writeInternal(offset, data, size, handle);
 }
 
-bool Device::write(uint64_t offset, Buffer buffer) {
+bool Device::write(uint64_t offset, Buffer buffer, int handle) {
   const auto size = buffer.size();
   XDCHECK_LE(offset + buffer.size(), size_);
   uint8_t* data = reinterpret_cast<uint8_t*>(buffer.data());
@@ -377,10 +405,13 @@ bool Device::write(uint64_t offset, Buffer buffer) {
       return false;
     }
   }
-  return writeInternal(offset, data, size);
+  return writeInternal(offset, data, size, handle);
 }
 
-bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size) {
+bool Device::writeInternal(uint64_t offset,
+                           const uint8_t* data,
+                           size_t size,
+                           int handle) {
   auto remainingSize = size;
   auto maxWriteSize = (maxWriteSize_ == 0) ? remainingSize : maxWriteSize_;
   bool result = true;
@@ -390,7 +421,7 @@ bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size) {
     XDCHECK_EQ(writeSize % ioAlignmentSize_, 0ul);
 
     auto timeBegin = getSteadyClock();
-    result = writeImpl(offset, writeSize, data);
+    result = writeImpl(offset, writeSize, data, handle);
     writeLatencyEstimator_.trackValue(
         toMicros((getSteadyClock() - timeBegin)).count());
 
@@ -418,18 +449,31 @@ bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size) {
 // returns true if successful, false otherwise.
 bool Device::readInternal(uint64_t offset, uint32_t size, void* value) {
   XDCHECK_EQ(reinterpret_cast<uint64_t>(value) % ioAlignmentSize_, 0ul);
-  XDCHECK_EQ(offset % ioAlignmentSize_, 0ul);
-  XDCHECK_EQ(size % ioAlignmentSize_, 0ul);
   XDCHECK_LE(offset + size, size_);
-  auto timeBegin = getSteadyClock();
-  bool result = readImpl(offset, size, value);
-  readLatencyEstimator_.trackValue(
-      toMicros(getSteadyClock() - timeBegin).count());
-  if (!result) {
-    readIOErrors_.inc();
-    return result;
+  uint8_t* data = reinterpret_cast<uint8_t*>(value);
+  auto remainingSize = size;
+  auto maxReadSize = (maxIOSize_ == 0) ? remainingSize : maxIOSize_;
+  bool result = true;
+  uint64_t curOffset = offset;
+  while (remainingSize > 0) {
+    auto readSize = std::min<size_t>(maxReadSize, remainingSize);
+    XDCHECK_EQ(curOffset % ioAlignmentSize_, 0ul);
+    XDCHECK_EQ(size % ioAlignmentSize_, 0ul);
+
+    auto timeBegin = getSteadyClock();
+    result = readImpl(curOffset, readSize, data);
+    readLatencyEstimator_.trackValue(
+        toMicros(getSteadyClock() - timeBegin).count());
+
+    if (!result) {
+      readIOErrors_.inc();
+      return false;
+    }
+    bytesRead_.add(readSize);
+    curOffset += readSize;
+    data += readSize;
+    remainingSize -= readSize;
   }
-  bytesRead_.add(size);
   if (encryptor_) {
     XCHECK_EQ(offset % encryptor_->encryptionBlockSize(), 0ul);
     auto res = encryptor_->decrypt(
@@ -440,7 +484,7 @@ bool Device::readInternal(uint64_t offset, uint32_t size, void* value) {
       return false;
     }
   }
-  return true;
+  return result;
 }
 
 // This API reads size bytes from the Device from offset into a Buffer and
@@ -504,15 +548,15 @@ std::string IOOp::toString() const {
       offset_, size_, data_, resubmitted_);
 }
 
-bool IOOp::done(ssize_t status) {
+bool IOOp::done(ssize_t size) {
   XDCHECK(parent_.opType_ == READ || parent_.opType_ == WRITE);
 
-  bool result = (status == size_);
+  bool result = (size == size_);
   if (!result) {
     // Report IO errors
     XLOG_N_PER_MS(ERR, 10, 1000) << folly::sformat(
         "[{}] IO error: {} ret={} errno={} ({})", parent_.context_.getName(),
-        toString(), status, errno, std::strerror(errno));
+        toString(), size, errno, std::strerror(errno));
   }
 
   // Check for timeout
@@ -540,12 +584,14 @@ IOReq::IOReq(IoContext& context,
              OpType opType,
              uint64_t offset,
              uint32_t size,
-             void* data)
+             void* data,
+             int handle)
     : context_(context),
       opType_(opType),
       offset_(offset),
       size_(size),
-      data_(data) {
+      data_(data),
+      handle_(handle) {
   uint8_t* buf = reinterpret_cast<uint8_t*>(data_);
   uint32_t idx = 0;
   if (fvec.size() > 1) {
@@ -559,14 +605,15 @@ IOReq::IOReq(IoContext& context,
 
       ops_.emplace_back(*this, idx++, fvec[fdIdx].fd(),
                         stripeStartOffset + ioOffsetInStripe, allowedIOSize,
-                        buf);
+                        buf, handle_);
 
       size -= allowedIOSize;
       offset += allowedIOSize;
       buf += allowedIOSize;
     }
   } else {
-    ops_.emplace_back(*this, idx++, fvec[0].fd(), offset_, size_, data_);
+    ops_.emplace_back(*this, idx++, fvec[0].fd(), offset_, size_, data_,
+                      handle_);
   }
 
   numRemaining_ = ops_.size();
@@ -632,7 +679,7 @@ std::shared_ptr<IOReq> IoContext::submitRead(
     uint32_t size,
     void* data) {
   auto req = std::make_shared<IOReq>(*this, fvec, stripeSize, OpType::READ,
-                                     offset, size, data);
+                                     offset, size, data, -1);
   submitReq(req);
   return req;
 }
@@ -642,9 +689,11 @@ std::shared_ptr<IOReq> IoContext::submitWrite(
     uint32_t stripeSize,
     uint64_t offset,
     uint32_t size,
-    const void* data) {
-  auto req = std::make_shared<IOReq>(*this, fvec, stripeSize, OpType::WRITE,
-                                     offset, size, const_cast<void*>(data));
+    const void* data,
+    int handle) {
+  auto req =
+      std::make_shared<IOReq>(*this, fvec, stripeSize, OpType::WRITE, offset,
+                              size, const_cast<void*>(data), handle);
   submitReq(req);
   return req;
 }
@@ -700,14 +749,16 @@ AsyncIoContext::AsyncIoContext(std::unique_ptr<folly::AsyncBase>&& asyncBase,
                                size_t id,
                                folly::EventBase* evb,
                                size_t capacity,
-                               bool useIoUring)
+                               bool useIoUring,
+                               std::vector<std::shared_ptr<FdpNvme>> fdpNvmeVec)
     : asyncBase_(std::move(asyncBase)),
       id_(id),
       qDepth_(capacity),
-      useIoUring_(useIoUring) {
+      useIoUring_(useIoUring),
+      fdpNvmeVec_(fdpNvmeVec) {
 #ifdef CACHELIB_IOURING_DISABLE
   // io_uring is not available on the system
-  XDCHECK(!useIoUring_);
+  XDCHECK(!useIoUring_ && !(fdpNvmeVec_.size() > 0));
   useIoUring_ = false;
 #endif
   if (evb) {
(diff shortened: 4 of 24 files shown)


cachelib/navy/common/Device.cpp
```diff
@@ -234,6 +242,14 @@ class AsyncIoContext : public IoContext {
  private:
   void handleCompletion(folly::Range<folly::AsyncBaseOp**>& completed);
 
+  std::unique_ptr<folly::AsyncBaseOp>
+   prepAsyncIo(int fd, uint8_t opType, void* data,
+          uint32_t size, uint64_t offset, void* userdata, bool useIoUring);
```
jaesoo-fb, line 247: nit. remove `useIoUring`; it is a member variable.

cachelib/navy/common/Device.cpp
```diff
@@ -234,6 +242,14 @@ class AsyncIoContext : public IoContext {
...
+
+  std::unique_ptr<folly::AsyncBaseOp>
+    prepUringCmdAsyncIo(int fd, uint8_t opType, void* data,
+      uint32_t size, uint64_t offset, void* userdata, int handle);
```
jaesoo-fb, line 251: nit. just `prepNvmeIo` or so. Also, can we make this being called inside `prepAsyncIo`?

cachelib/navy/common/Device.cpp
```diff
@@ -310,6 +334,11 @@ class FileDevice : public Device {
   // determine the capacity of an io_uring/libaio queue
   const uint32_t qDepthPerContext_;
 
+  // Whether FDP Placement mode is enabled or not
+  bool isFDPEnabled_{false};
```
jaesoo-fb, line 338: This looks not needed. Just `!!fdpNvme_`.

cachelib/navy/common/Device.cpp
```diff
@@ -352,19 +389,27 @@ class MemoryDevice final : public Device {
 } // namespace
 
 bool Device::write(uint64_t offset, BufferView view) {
+  return write(offset, view, -1);
+}
+
+bool Device::write(uint64_t offset, BufferView view, int handle) {
```
jaesoo-fb, line 395: Maybe just define `write` as `write(..., int handle = -1)`?

cachelib/navy/common/Device.cpp
```diff
@@ -352,19 +389,27 @@ class MemoryDevice final : public Device {
...
   const uint8_t* data = reinterpret_cast<const uint8_t*>(view.data());
-  return writeInternal(offset, data, size);
+  return writeInternal(offset, data, size, handle);
 }
 
 bool Device::write(uint64_t offset, Buffer buffer) {
+  return write(offset, std::move(buffer), -1);
+}
+
+bool Device::write(uint64_t offset, Buffer buffer, int handle) {
```
jaesoo-fb, line 412: Same. Maybe just define write as write(..., int handle = -1)?

cachelib/navy/common/Device.cpp
```diff
@@ -377,20 +422,22 @@ bool Device::write(uint64_t offset, Buffer buffer) {
...
-  return writeInternal(offset, data, size);
+  return writeInternal(offset, data, size, handle);
 }
 
-bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size) {
+bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size,
+          int handle) {
   auto remainingSize = size;
   auto maxWriteSize = (maxWriteSize_ == 0) ? remainingSize : maxWriteSize_;
+  maxWriteSize = (maxIOSize_ == 0) ? maxWriteSize : maxIOSize_;
```
jaesoo-fb, line 432: remove this after fix setting `maxIOSize_`.

cachelib/navy/common/Device.cpp
```diff
@@ -863,6 +981,15 @@ FileDevice::FileDevice(std::vector<folly::File>&& fvec,
   // for the path where device IO is called from non-fiber thread
   // (e.g., recovery path, read random alloc path)
   syncIoContext_ = std::make_unique<SyncIoContext>();
+#ifndef CACHELIB_IOURING_DISABLE
+  if (isFDPEnabled_) {
+    fdpNvme_ = std::make_shared<FdpNvme>(fvec_[0].fd());
+    setMaxIOSize(fdpNvme_->getMaxTfrSize());
```
jaesoo-fb, line 987: Fix `maxWriteSize_` too.
```
maxIOSize_ = fdpNvme_->getMaxTfrSize();
maxWriteSize_ = min(maxWriteSize_, maxIOSize_);
```

cachelib/navy/common/Device.cpp
```diff
@@ -418,18 +465,30 @@ bool Device::writeInternal(uint64_t offset, const uint8_t* data, size_t size) {
...
+
+    auto timeBegin = getSteadyClock();
+    result = readImpl(offset, readSize, data);
+    readLatencyEstimator_.trackValue(
+        toMicros(getSteadyClock() - timeBegin).count());
+
+    if (result) {
+      bytesRead_.add(readSize);
+    } else {
+      break;
```
jaesoo-fb, line 486: Can we just return here?
```
readIOErrors_.inc(); return false;
```

cachelib/navy/common/Device.cpp
```diff
@@ -762,7 +837,7 @@ void AsyncIoContext::handleCompletion(
                          aop->result(), iop->toString());
     }
 
-    iop->done(aop->result());
+    iop->done(aop->result(), fdpNvme_ != nullptr);
```
jaesoo-fb, line 840:
```
auto result = aop->result();
if (fdpNvme_) {
  // 0 means success, so get completed size from iop
  result = !result? iop->size_ : 0;  
}
```

cachelib/navy/common/Device.cpp
```diff
@@ -504,10 +566,14 @@ std::string IOOp::toString() const {
       offset_, size_, data_, resubmitted_);
 }
 
-bool IOOp::done(ssize_t status) {
+bool IOOp::done(ssize_t status, bool useIoUringCmd) {
```
jaesoo-fb, line 569: Can you keep prev one and change `status` to `size` or `len`?

jaesoo-fb (commented):
Can you check if build is OK with removing liburing package?

cachelib/navy/common/Device.cpp
```diff
@@ -905,6 +1038,7 @@ IoContext* FileDevice::getIoContext() {
 
     // Create new context if on event base thread or useIoUring_ is enabled
     bool useIoUring = ioEngine_ == IoEngine::IoUring;
+    bool useFdpIoUringCmd = isFDPEnabled_;
```
jaesoo-fb, line 1041: Why additional variable?

cachelib/navy/common/Device.cpp
```diff
@@ -863,6 +981,15 @@ FileDevice::FileDevice(std::vector<folly::File>&& fvec,
   // for the path where device IO is called from non-fiber thread
   // (e.g., recovery path, read random alloc path)
   syncIoContext_ = std::make_unique<SyncIoContext>();
+#ifndef CACHELIB_IOURING_DISABLE
+  if (isFDPEnabled_) {
+    fdpNvme_ = std::make_shared<FdpNvme>(fvec_[0].fd());
```
jaesoo-fb, line 986: How do we guarantee `fdpNvme_` is initialized successfully? If FDP is not enabled, we should reset it.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+ * limitations under the License.
+ */
+
+#include "cachelib/navy/common/FdpNvme.h"
+
+#include <linux/nvme_ioctl.h>
+#include <sys/ioctl.h>
+#include <errno.h>
+#include <cstring>
+#include <numeric>
```
jaesoo-fb, line 23: nit. in alphabetical order?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+  // Get NS specific Fdp Placement Handle(PHNDL)
+  if (nextPIDIdx_ <= maxPIDIdx_) {
+    phndl = nextPIDIdx_++;
+  } else {
+    phndl = kDefaultPIDIdx;
+  }
+  // Get Device specific Fdp Placement ID for PHNDL
+  auto pid = getFdpPID(phndl);
+
+  return static_cast<int>(pid);
```
jaesoo-fb, line 47: Do we need to expose the placement id here? Why not just give out the `phndl`?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+int FdpNvme::allocateFdpHandle() {
+  uint16_t phndl;
+
+  // Get NS specific Fdp Placement Handle(PHNDL)
+  if (nextPIDIdx_ <= maxPIDIdx_) {
+    phndl = nextPIDIdx_++;
+  } else {
+    phndl = kDefaultPIDIdx;
```
jaesoo-fb, line 42: Can we print some log here?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+#include <errno.h>
+#include <cstring>
+#include <numeric>
+
+namespace facebook {
+namespace cachelib {
+namespace navy {
+
+FdpNvme::FdpNvme(int fd) {
+  nvmeData_ = readNvmeInfo(fd);
```
jaesoo-fb, line 30: Can you put this initialize in `initializeFDP` too?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+    phndl = kDefaultPIDIdx;
+  }
+  // Get Device specific Fdp Placement ID for PHNDL
+  auto pid = getFdpPID(phndl);
+
+  return static_cast<int>(pid);
+}
+
+void FdpNvme::initializeFDP(int fd) {
+  nextPIDIdx_ = kDefaultPIDIdx + 1;
```
jaesoo-fb, line 51: nit. Default initialize this or put in the contructor's initialize list.

cachelib/navy/common/FdpNvme.h
```diff
@@ -0,0 +1,271 @@
...
+
+  // Generic NVMe IO mgmnt receive cmd
+  int nvmeIOMgmtRecv(int fd, uint32_t nsid, void *data, uint32_t data_len,
+      uint16_t mos, uint8_t mo);
+
+  // 0u is considered as the default placement ID
+  static constexpr uint16_t kDefaultPIDIdx = 0u;
+
+  // The mapping table of <PHNDL, PID> of Namespaces
+  std::map<uint16_t, uint16_t> placementIDs_{};
```
jaesoo-fb, line 263: Why map? Just `std::vector<uint16_t>` would work.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+  Buffer buffer = nvmeFdpStatus(fd);
+  if (!buffer.isNull()) {
```
jaesoo-fb, line 54: Raise exception, so that `Device` code can handle and disable fdp?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+    if (ruh_status->nruhsd) {
+      maxPIDIdx_ = ruh_status->nruhsd - 1;
+      for (uint16_t i = 0; i <= maxPIDIdx_; ++i) {
+        placementIDs_[i] = ruh_status->ruhss[i].pid;
+      }
+    }
+  }
+  XLOG(INFO)<< "Creating NvmeFdp, fd :"<< fd<<" Num of PID: "
+            <<maxPIDIdx_ + 1<<" First PID: "<<placementIDs_[0]
+            <<" Last PID: "<<placementIDs_[maxPIDIdx_];
```
jaesoo-fb, line 67: This looks like a debug info. `XLOG(DEBUG)` or so. Also, can you use formatted string form? i.e., `XLOGF(DEBUG, ...)`?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+}
+
+int FdpNvme::nvmeIOMgmtRecv(int fd, uint32_t nsid, void *data,
+                  uint32_t data_len, uint16_t mos, uint8_t mo) {
+  uint32_t cdw10 = (mo & 0xf) | (mos & 0xff << 16);
+  uint32_t cdw11 = (data_len >> 2) - 1;
```
jaesoo-fb, line 73: Can you add some general comment? like `MOS`, `MO`, cdw11 is 0-based, etc.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+      }
+    }
+  }
+  XLOG(INFO)<< "Creating NvmeFdp, fd :"<< fd<<" Num of PID: "
+            <<maxPIDIdx_ + 1<<" First PID: "<<placementIDs_[0]
+            <<" Last PID: "<<placementIDs_[maxPIDIdx_];
+}
+
+int FdpNvme::nvmeIOMgmtRecv(int fd, uint32_t nsid, void *data,
+                  uint32_t data_len, uint16_t mos, uint8_t mo) {
```
jaesoo-fb, line 71: Can you change the order of mos and mo? `uint8_t op, uint16_t op_specific`?

cachelib/navy/common/FdpNvme.h
```diff
@@ -0,0 +1,271 @@
...
+
+  // Prepares the Uring_Cmd sqe for write command with FDP handle.
+  void prepWriteUringCmdSqe(struct io_uring_sqe& sqe, int fd, void* buf,
+      size_t size, off_t start, int handle);
+
+ private:
+  // Prepares the Uring_Cmd sqe for read/write command with FDP directives.
+  void prepFdpUringCmdSqe(struct io_uring_sqe& sqe, int fd, void* buf,
+      size_t size, off_t start, uint8_t opcode, uint8_t dtype,
+      uint16_t dspec, NvmeData& nvmeData);
```
jaesoo-fb, line 241: Why do we need to get `nvmeData` as parameter here?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+    .timeout_ms     = NVME_DEFAULT_IOCTL_TIMEOUT,
+  };
+
+  err = ioctl(fd, NVME_IOCTL_ADMIN_CMD, &cmd);
+  if (err) {
+    XLOG(ERR)<< "failed to fetch identify namespace";
+    return NvmeData{};
+  }
+
+  auto lbaShift = (uint32_t)ilog2(1 << ns.lbaf[(ns.flbas & 0x0f)].ds);
```
jaesoo-fb, line 220: The `ds` should be in terms of power of two already?

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,228 @@
...
+  XLOG(INFO) <<"Nvme Device Info, NS Id: " <<namespace_id<<", lbashift: "
+             <<lbaShift<<", size: "<<ns.nsze<<", Max Tfr size: "<<maxTfrSize;
```
jaesoo-fb, line 222: nit. Can you do the formating? and represent the size as byte as well?

jaesoo-fb (commented):
Can you check if this compiles in supported platforms here https://github.com/facebook/CacheLib/actions?

arungeorge83:
> Can you check if build is OK with removing liburing package?

FdpNvme.cpp is not protected for this case. I guess you want to bring them under -DCACHELIB_IOURING_DISABLE, right?

cachelib/allocator/nvmcache/NavyConfig.h
```diff
@@ -678,6 +682,8 @@ class NavyConfig {
   // Whether to use write size (instead of parcel size) for Navy admission
   // policy.
   bool useEstimatedWriteSize_{false};
+  // Whether Navy will use the FDP data placement directives or not.
```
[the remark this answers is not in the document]
arungeorge83: sure, will do it.

cachelib/navy/bighash/BigHash.cpp
```diff
@@ -95,6 +95,7 @@ BigHash::BigHash(Config&& config, ValidConfigTag)
         bucketSize_,
         cacheBaseOffset_);
   reset();
+  placementHandle_ = device_.allocatePlacementHandle();
```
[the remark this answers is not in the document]
arungeorge83: yes, sure

cachelib/navy/common/Device.cpp
arungeorge83: I thought of readability; to emphasize why we are taking the io_uring_cmd path. Anyway it can be removed.

cachelib/navy/common/Device.cpp
arungeorge83: it was initialized to nullptr.
std::shared_ptr<FdpNvme> fdpNvme_{nullptr};

arungeorge83:
> Can you check if this compiles in supported platforms here https://github.com/facebook/CacheLib/actions?
As per the status,
'This workflow is awaiting approval from a maintainer in https://github.com/facebook/CacheLib/pull/277'
Could you check it and approve?
Or should I verify it locally?

cachelib/navy/common/FdpNvme.cpp
arungeorge83: Yes, possible.

cachelib/navy/common/FdpNvme.h
arungeorge83: yes, if we are returning phndl instead of pid. Thanks for pointing out that.

jaesoo-fb:
> As per the status, 'This workflow is awaiting approval from a maintainer in #277' Could you check it and approve? Or should I verify it locally?

Oh, my bad. I just started the workflows. I expect some of them to be failed, so please fix it as it appears.

cachelib/navy/common/Device.cpp
jaesoo-fb: I meant, what if the constructor fails to initialize FDP NVMe.

cachelib/navy/Factory.cpp
```diff
@@ -460,7 +481,11 @@ std::unique_ptr<Device> createFileDevice(
   for (const auto& path : filePaths) {
     folly::File f;
     try {
-      f = openCacheFile(path, fdSize, truncateFile);
+      if(isFDPEnabled) {
+        f = openNvmeCharFile(path);
+      } else {
+        f = openCacheFile(path, fdSize, truncateFile);
+      }
```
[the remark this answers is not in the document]
arungeorge83: IIUC, your requirement is as follows,
There are partitions like nvmeXnYp1, nvmeXnYp2 ... , and one of them is used in this cachelib instance.
(Ex: nvmCachePath = "/dev/nvmeXnYp2")
OR is it just one partition in that namespace, but it happened to be partition-formatted ?

>> The rationale is that this will allow us to apply the size check and fallocate thing the same
Is this steps used for device simulation scenarios?

Anyway the proposed approach is doable and making sense.

cachelib/navy/common/FdpNvme.cpp
arungeorge83: yes, you are right. will correct it

cachelib/navy/common/Device.cpp
```diff
@@ -700,16 +748,19 @@ AsyncIoContext::AsyncIoContext(std::unique_ptr<folly::AsyncBase>&& asyncBase,
...
+                               std::shared_ptr<FdpNvme> fdpNvme)
     : asyncBase_(std::move(asyncBase)),
       id_(id),
       qDepth_(capacity),
-      useIoUring_(useIoUring) {
+      useIoUring_(useIoUring),
+      fdpNvme_(fdpNvme) {
 #ifdef CACHELIB_IOURING_DISABLE
   // io_uring is not available on the system
   XDCHECK(!useIoUring_);
```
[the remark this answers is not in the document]
jaesoo-fb: If `CACHELIB_IOURING_DISABLE`, both of `useIoUring_` and `fdbNvme_ ` should be nullptr, right?

cachelib/navy/common/Device.cpp
```diff
@@ -863,6 +955,25 @@ FileDevice::FileDevice(std::vector<folly::File>&& fvec,
...
+    }
+
+    if (!!fdpNvme_) {
+      auto maxIOSize = fdpNvme_->getMaxTfrSize();
+      if (maxIOSize) {
+        setMaxIOSize(maxIOSize);
+        maxDeviceWriteSize = std::min<size_t>(maxDeviceWriteSize, maxIOSize);
+        setMaxWriteSize(maxDeviceWriteSize);
+      }
+    }
```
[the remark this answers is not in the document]
jaesoo-fb: Yeah, something like that. Both `maxIOSize_` and `maxDeviceWriteSize_` would be set in the constructor initializer list.

cachelib/navy/Factory.cpp
```diff
@@ -439,6 +460,26 @@ std::unique_ptr<AbstractCache> createCache(std::unique_ptr<CacheProto> proto) {
...
+  if (!isValidNvmeDevice(blockDevice)) {
+    throw std::invalid_argument("Invalid NVMe device name");
+  }
+
+  // Extract dev and NS IDs, and ignore partition ID.
+  // Example: extract the string '0n1' from '/dev/nvme0n1p1'
+  size_t devPos = blockDevice.find_first_of("0123456789");
+  size_t pPos = blockDevice.find('p', devPos);
+
+  return "/dev/ng" + blockDevice.substr(devPos, pPos - devPos);
```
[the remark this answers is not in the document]
jaesoo-fb: Actually, I think it would be good to have all nvme specific code in `Device.cpp` or `FdpNvme.h`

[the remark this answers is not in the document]
arungeorge83: okay,sure.

>>  So, I think we can restrict FdpNvme is only supported for single block device only for now.
Yes, I am adding explicit checks and exceptions to limit the FDP support for single device for now.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,321 @@
...
+  size_t pPos = blockDevice.find('p', bPos);
+
+  XDCHECK(bPos < std::string::npos);
+  return blockDevice.substr(bPos, pPos - bPos);
+}
+
+// It returns nvme0n1p1 for /dev/nvme0n1p1 (partitioned NS),
+// and returns "" for /dev/nvme0n1 (non-partitioned NS)
+std::string getNvmePartName(const std::string& blockDevice) {
+  size_t bPos = blockDevice.find_last_of("/") + 1;
```
jaesoo-fb, line 207: I think you need to check the result of `find_last_of`. `npos + 1` would be 0, which make below code not work as intended.

cachelib/navy/common/tests/DeviceTest.cpp
```diff
@@ -213,12 +213,14 @@ struct DeviceParamTest
       uint32_t maxDeviceWriteSize,
       std::shared_ptr<DeviceEncryptor> encryptor) {
     device_ = createDirectIoFileDevice(std::move(fVec),
+                                       std::move(std::vector<std::string>()),
```
jaesoo-fb, line 216: nit. just "{}" for the null vector.

cachelib/navy/common/Device.cpp
```diff
@@ -310,6 +332,9 @@ class FileDevice : public Device {
   // determine the capacity of an io_uring/libaio queue
   const uint32_t qDepthPerContext_;
 
+  // Device info vector for FDP support
+  const std::vector<std::shared_ptr<FdpNvme>> fdpNvmeVec_{};
+
```
jaesoo-fb, line 337: Could you move this to right below `fvec_` to match with the initialization order?

cachelib/navy/common/Device.cpp
```diff
@@ -331,7 +356,8 @@ class MemoryDevice final : public Device {
  private:
   bool writeImpl(uint64_t offset,
                  uint32_t size,
-                 const void* value) noexcept override {
+                 const void* value,
+                 int handle) noexcept override {
```
jaesoo-fb, line 360: nit. `int /* unused */`

cachelib/navy/common/Device.cpp
```diff
@@ -797,46 +879,60 @@ bool AsyncIoContext::submitIo(IOOp& op) {
...
     asyncOp->pread(op.fd_, op.data_, op.size_, op.offset_);
   } else {
     XDCHECK_EQ(req.opType_, OpType::WRITE);
-    asyncOp->setUserData(&op);
     asyncOp->pwrite(op.fd_, op.data_, op.size_, op.offset_);
   }
 
-  asyncBase_->submit(asyncOp.release());
-  op.submitTime_ = getSteadyClock();
+  return std::move(asyncOp);
```
jaesoo-fb, line 888: nit. `return asyncOp;`

cachelib/navy/common/Device.cpp
```diff
@@ -987,12 +1149,14 @@ std::unique_ptr<Device> createDirectIoFileDevice(
     uint32_t maxDeviceWriteSize,
     std::shared_ptr<DeviceEncryptor> encryptor) {
   return createDirectIoFileDevice(std::move(fVec),
+                                  std::move(std::vector<std::string>()),
```
jaesoo-fb, line 1152: nit. Just "{}"

cachelib/navy/common/FdpNvme.h
```diff
@@ -0,0 +1,216 @@
...
+#endif
+
+// FDP specific info and handling
+//
+// This embeds the FDP semantics and specific io-handling.
+// Note: IO with FDP semantics need to be sent through Io_Uring_cmd interface
+// as of now; and not supported through conventional block interfaces.
+class FdpNvme {
+ public:
+  FdpNvme(const std::string& fileName);
```
jaesoo-fb, line 137: Make this explicit, i.e., `explicit FdpNvme(...)`

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,321 @@
...
+ */
+
+#include "cachelib/navy/common/FdpNvme.h"
+
+#include <errno.h>
+#include <linux/nvme_ioctl.h>
+#include <sys/ioctl.h>
+
+#include <cstring>
+#include <numeric>
```
jaesoo-fb, line 24: Is this used?

cachelib/navy/common/Device.cpp
```diff
@@ -788,6 +844,32 @@ bool AsyncIoContext::submitIo(IOOp& op) {
   }
```
jaesoo-fb, line 844: I cannot add comment above. Could you remove `req` which is not used.

```
@@ -773,9 +829,8 @@
 }

 bool AsyncIoContext::submitIo(IOOp& op) {
-  IOReq& req = op.parent_;
-
   op.startTime_ = getSteadyClock();
+
   while (numOutstanding_ >= qDepth_) {
     if (qDepth_ > 1) {
       XLOG_EVERY_MS(ERR, 10000) << fmt::format(
```

cachelib/navy/testing/MockDevice.h
```diff
@@ -44,6 +44,14 @@ class MockDevice : public Device {
   MOCK_METHOD3(readImpl, bool(uint64_t, uint32_t, void*));
   MOCK_METHOD3(writeImpl, bool(uint64_t, uint32_t, const void*));
   MOCK_METHOD0(flushImpl, void());
+  MOCK_METHOD0(allocatePlacementHandle, int());
+
+  virtual bool writeImpl(uint64_t offset,
```
jaesoo-fb, line 49: remove `virtual` here and add `override` instead. i.e., `bool writeImpl(...) override {`

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,321 @@
...
+  return blockDevice.substr(bPos, pPos - bPos);
+}
+
+// It returns nvme0n1p1 for /dev/nvme0n1p1 (partitioned NS),
+// and returns "" for /dev/nvme0n1 (non-partitioned NS)
+std::string getNvmePartName(const std::string& blockDevice) {
+  size_t bPos = blockDevice.find_last_of("/") + 1;
+  size_t pPos = blockDevice.find('p', bPos);
+
+  XDCHECK(bPos < std::string::npos);
```
jaesoo-fb, line 210: Remove this.

jaesoo-fb (commented):
Overall looks good. Thanks for the work.

Could you apply below patch?
```
diff --git a/fbcode/cachelib/navy/common/TARGETS b/fbcode/cachelib/navy/common/TARGETS
--- a/fbcode/cachelib/navy/common/TARGETS
+++ b/fbcode/cachelib/navy/common/TARGETS
@@ -7,6 +7,7 @@
     srcs = [
         "Buffer.cpp",
         "Device.cpp",
+        "FdpNvme.cpp",
         "Hash.cpp",
         "SizeDistribution.cpp",
         "Types.cpp",
@@ -15,6 +16,7 @@
         "Buffer.h",
         "CompilerUtils.h",
         "Device.h",
+        "FdpNvme.h",
         "Hash.h",
         "NavyThread.h",
         "SizeDistribution.h",
@@ -27,7 +29,6 @@
         "//folly:function",
         "//folly:thread_local",
         "//folly/experimental/io:async_io",
-        "//folly/experimental/io:io_uring",
         "//folly/hash:checksum",
         "//folly/hash:hash",
         "//folly/io/async:event_base_manager",
@@ -41,6 +42,8 @@
         "//folly:file",
         "//folly:portability",
         "//folly:range",
+        "//folly/experimental/io:async_base",
+        "//folly/experimental/io:io_uring",
         "//folly/fibers:core_manager",
         "//folly/fibers:fiber_manager_map",
         "//folly/fibers:timed_mutex",
@@ -50,4 +53,7 @@
         "//folly/lang:bits",
         "//folly/logging:logging",
     ],
+    exported_external_deps = [
+        ("liburing", None, "uring"),
+    ],
 )
 ```

**Comment by arungeorge83** [comment]
> Overall looks good. Thanks for the work.
> 
> Could you apply below patch?
> 
> ```
> diff --git a/fbcode/cachelib/navy/common/TARGETS b/fbcode/cachelib/navy/common/TARGETS
> --- a/fbcode/cachelib/navy/common/TARGETS
> +++ b/fbcode/cachelib/navy/common/TARGETS
> @@ -7,6 +7,7 @@
>      srcs = [
>          "Buffer.cpp",
>          "Device.cpp",
> +        "FdpNvme.cpp",
>          "Hash.cpp",
>          "SizeDistribution.cpp",
>          "Types.cpp",
> @@ -15,6 +16,7 @@
>          "Buffer.h",
>          "CompilerUtils.h",
>          "Device.h",
> +        "FdpNvme.h",
>          "Hash.h",
>          "NavyThread.h",
>          "SizeDistribution.h",
> @@ -27,7 +29,6 @@
>          "//folly:function",
>          "//folly:thread_local",
>          "//folly/experimental/io:async_io",
> -        "//folly/experimental/io:io_uring",
>          "//folly/hash:checksum",
>          "//folly/hash:hash",
>          "//folly/io/async:event_base_manager",
> @@ -41,6 +42,8 @@
>          "//folly:file",
>          "//folly:portability",
>          "//folly:range",
> +        "//folly/experimental/io:async_base",
> +        "//folly/experimental/io:io_uring",
>          "//folly/fibers:core_manager",
>          "//folly/fibers:fiber_manager_map",
>          "//folly/fibers:timed_mutex",
> @@ -50,4 +53,7 @@
>          "//folly/lang:bits",
>          "//folly/logging:logging",
>      ],
> +    exported_external_deps = [
> +        ("liburing", None, "uring"),
> +    ],
>  )
> ```

I could not find the file TARGETS. Is it something not present in open source version?

File: cachelib/navy/common/FdpNvme.cpp [hunk 43]
```diff
@@ -0,0 +1,321 @@
...
+                     pid);
+}
+
+// It returns "nvme0n1" for both "/dev/nvme0n1" and "/dev/nvme0n1p1"
+std::string getNvmeBlockName(const std::string& blockDevice) {
+  size_t bPos = blockDevice.find_last_of("/") + 1;
+  size_t pPos = blockDevice.find('p', bPos);
+
+  XDCHECK(bPos < std::string::npos);
+  return blockDevice.substr(bPos, pPos - bPos);
```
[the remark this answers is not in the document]
arungeorge83: I guess, the result of this line might have been compiler dependent. (Actually i did not assume partition :). will modify it)

jaesoo-fb:
> I could not find the file TARGETS. Is it something not present in open source version?

Yeah, this is for internal build. Please ignore.

cachelib/navy/testing/MockDevice.h
arungeorge83: This results in compilation error (due to gtest restrictions on default argument - int handle - I guess).
How about adding a line ?
MOCK_METHOD4(writeImpl, bool(uint64_t, uint32_t, const void*, int));

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -0,0 +1,321 @@
...
+
+    return NvmeData{namespace_id, lbaShift, maxTfrSize, startLba};
+  } catch (const std::system_error& e) {
+    XLOGF(ERR, "Exception in openNvmeCharFile for: {}", blockDevice);
+    throw;
+  }
+}
+
+bool isValidNvmeDevice(const std::string& fileName) {
+  return std::regex_match(fileName, std::regex("^/dev/nvme\\d+n\\d+(p\\d+)?$"));
```
jaesoo-fb, line 280: Could you replace this with the equivalent (`re2::RE2::FullMatch`?) from RE2? https://github.com/google/re2 

We cannot use `regex_match` due to its vulnerability to segfault https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61582&fbclid=IwAR16Xwiru8QryYObTnoATM5WwCSTX9IRVorMwDxYq9E0sxrsBQTw-H1iKH4

cachelib/navy/common/FdpNvme.h
```diff
@@ -0,0 +1,216 @@
...
+#include <folly/File.h>
+#include <folly/experimental/io/AsyncBase.h>
+#include <folly/experimental/io/IoUring.h>
+
+#include "cachelib/navy/common/Buffer.h"
+#include "cachelib/navy/common/Device.h"
+
+#ifndef CACHELIB_IOURING_DISABLE
+#include <liburing.h>
+#endif
```
jaesoo-fb, line 28: FYI. Actually, we should not include `liburing.h` directly here due to some internal dependency issues causing test failures on some targets. I will remove this.

cachelib/navy/common/Device.cpp
```diff
@@ -923,8 +1025,16 @@ IoContext* FileDevice::getIoContext() {
     std::unique_ptr<folly::AsyncBase> asyncBase;
     if (useIoUring) {
 #ifndef CACHELIB_IOURING_DISABLE
-      asyncBase = std::make_unique<folly::IoUring>(qDepthPerContext_, pollMode,
-                                                   qDepthPerContext_);
+      if (fdpNvmeVec_.size() > 0) {
+        folly::IoUringOp::Options options;
+        options.sqe128 = true;
+        options.cqe32 = true;
```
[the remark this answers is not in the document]
arungeorge83: right, we can remove the line "options.cqe32 = true;" for now. We don't have the big_cqe use cases in the current code.

cachelib/navy/common/FdpNvme.h
arungeorge83: Ok. 
Then should we include <liburing.h> through <folly/experimental/io/IoUring.h> which include <liburing.h> in the file?

cachelib/navy/common/Device.cpp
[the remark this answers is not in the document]
jaesoo-fb: @arungeorge83 Actually, I would like to remove this option completely. So, we don't need `sqe128`, either, right?

cachelib/navy/common/FdpNvme.h
jaesoo-fb: @arungeorge83 Yeah, right. It will include the `liburing.h` from the repo instead of that installed on the system.

FYI, I am taking care of minor changes including those required for the internal build, so you don't need to update the commit.

cachelib/navy/common/Device.cpp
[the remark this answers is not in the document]
arungeorge83: @jaesoo-fb Actually both sqe128 and cqe32 are mandatory for NVMe passthrough commands. Else the kernel NVMe driver will reject those ios. 
Reference: https://elixir.bootlin.com/linux/latest/source/drivers/nvme/host/ioctl.c#L742, https://github.com/Iceber/iouring-go/issues/21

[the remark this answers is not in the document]
arungeorge83: Also from Kanchan Joshi's words (Author of the IO passthru patch - https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=456cba386e94f22fa1b1426303fdcac9e66b1417),
NVMe passthrough commands are of 72 bytes in size. This can not fit into regular SQE. Therefore big SQE (128 bytes) is required.
Also passthrough commands may return more than one result to userspace. Returning addtional result is not possible with regular CQE. Hence big CQE is needed.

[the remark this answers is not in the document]
jaesoo-fb: Thanks for clarification. Yeah, it looks clear here https://elixir.bootlin.com/linux/v6.7/source/drivers/nvme/host/ioctl.c#L742

jaesoo-fb:
Submitted with the fixup https://github.com/facebook/CacheLib/commit/b5d70a5f372f8b2950905e27c4adddf85c4bece7

PR #299, opened 2024-03-21
Add dev container settings

wankunde:
Add dev container settings, which are useful for local development.


jiayuebao:
@wankunde: Thanks for your interest in CacheLib. However, we would recommend just leaving the change locally instead of adding it to the CacheLib repo .

PR #307, opened 2024-04-22
Binary request generator/replayer

byrnedj:
This is the binary trace replayer/generator that we have been using to achieve max CPU utilization for the kvcache traces in cachebench. With this generator, we can achieve a throughput of over 20 million op/sec using kvcache workload in cachebench. As a comparison, using the CSV replay generator we see only ~1.6 million op/sec due to dynamic allocations and parsing overhead.

We avoid allocations by mmap'ing the request data into memory and using a Request pointer to point to the request data rather than allocating a new request wrapper for each request.

To generate a binary request file from an existing kvcache trace (using the "replay" generator).
1. Specify the kvcache trace name using the regular `traceFileNames` or `traceFileName` option. Specify other properties such as ampFactor too.
2. In the replayGeneratorConfig, specify `binaryFileName: "mybinaryfile.bin` as a config option
3. Run cachebench and wait for the binary file to be generated

To run a binary request trace specify the following:
1. Set generator to "binary-replay"
2. Set `traceFileName: "mybinaryfile.bin"` and set ampSizeFactor (if desired)

In summary - this patch offers much lower overhead of trace replaying. It does assumes the kvcache trace format and kvcache replay generator behavior. Additional features:
- fast forwarding of a trace
- preloading requests into memory
- object size amplification
- queue free for even lower request overhead

The limitations are:
- no trace amplification (however you can amplify the
original .csv trace and save it in binary format)
- ~4GB overhead per 100 million requests
- you need some disk space to store large traces

### DIFF
--- a/cachelib/cachebench/CMakeLists.txt
+++ b/cachelib/cachebench/CMakeLists.txt
@@ -41,6 +41,25 @@ target_link_libraries(cachelib_cachebench PUBLIC
   gflags
 )
 
+add_library (cachelib_binary_trace_gen
+  ./runner/Runner.cpp
+  ./runner/Stressor.cpp
+  ./util/CacheConfig.cpp
+  ./util/Config.cpp
+  ./workload/BlockChunkCache.cpp
+  ./workload/BlockChunkReplayGenerator.cpp
+  ./workload/PieceWiseCache.cpp
+  ./workload/OnlineGenerator.cpp
+  ./workload/WorkloadGenerator.cpp
+  ./workload/PieceWiseReplayGenerator.cpp
+  )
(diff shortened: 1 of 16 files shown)


therealgymmy (changes requested):
Thanks for sending this out! A few comments inline. :-)

therealgymmy:
@byrnedj: tested it out internally and verified 10x throughput improvement. The binary trace currently does not repeat if the specified operations are longer than the trace lenght, is this intended?

byrnedj:
I just added that functionality to the latest version.

therealgymmy:
Thanks let me re-import again.

PR #308, opened 2024-04-24
Adds documentation about FDP support in Cachelib

arungeorge83:
This adds the relevant documentation related to the FDP integration to CacheLib which was merged in commit 009e89b.

This commit adds a separate page containing all the information about FDP and its use in CacheLib. Additionally, this page introduces a potential user of CacheLib with all the relevant steps to setup up an FDP SSD and to run CacheLib with FDP enabled.

The parameter to enable FDP in CacheBench "deviceEnableFDP" has been updated to the CacheBench configuration page.

The parameter to enable FDP in Navy Config
"navyConfig.setEnableFDP(enableFDP)" is updated to the Hybrid Cache configuration page.

Information on the FdpNvme class is added to the Navy Architecture Guide page.


website/docs/Cache_Library_User_Guides/FDP_enabled_Cache.md
```diff
@@ -0,0 +1,92 @@
...
+                          void* buf,
+                          size_t size,
+                          off_t start,
+                          uint8_t opcode,
+                          uint8_t dtype,
+                          uint16_t dspec);
+```
+
+### When should FDP be enabled?
+When using Navy, FDP can help play a role in improving SSD endurance in comparison to Non-FDP. If the workload has both small objects and large objects, FDP's WAF gains will most likely be evident because FDP segregates these two in the SSD. In cases like CacheLib's CDN workload where BigHash is typically not configured, this FDP based segregation will make no difference to SSD WAF. On the other hand with CacheLib's KV Cache workload where both Big Hash and Block Cache are configured, we see the gains from using FDP.
```
therealgymmy, line 38: Can you add some numbers here from your experiments in KV Cache workloads FDP vs without?

roshanrn: Yes, we'll add the data we have from our KV Cache experiments with and without FDP.

roshanrn: We've updated the results in the commit. Please check and let us know.

therealgymmy: appreciate it!

roshanrn: With respect to configs:
We leave out configs in this description because we don't change the default configs obtained when downloading the traces. We expect the user to follow the steps mentioned in the "[Evaluating SSD hardware for Facebook workloads](https://cachelib.org/docs/Cache_Library_User_Guides/Cachebench_FB_HW_eval/)" page to [run the production traces with CacheBench](https://cachelib.org/docs/Cache_Library_User_Guides/Cachebench_FB_HW_eval/#running-cachebench-with-the-trace-workload). The only additional steps to enable FDP would be to use the flags mentioned in this page.

With respect to the specific trace used:
As mentioned the benefits of FDP can be seen with KV Cache traces and these benefits won't be seen with CDN or Storage traces because they don't configure the SOC.
There are multiple KV Cache traces in "s3://cachelib-workload-sharing/pub/kvcache/". Any of them can be used to observe the benefits. Again, this is why we don't point to a specific KV Cache trace in this description. 

However, we can update the PR to point to the particular KV Cache trace used in the results section presented for the sake of reproducibility.

haowu14 (approved):

Issue #309, opened 2024-04-28
Running simple-cache-example gives an error, flag 'v' was defined more than once


YuqingXiong:
**Describe the bug**
when i run the example "simple_cache", i find an error "ERROR: flag 'v' was defined more than once"

**To Reproduce**
1. `cd CacheLib/examples/simple_cache`
2. `./build.sh`
3. `./build/simple-cache-example`

**Expected behavior**
i want to run this example.

**Screenshots**
![image](https://github.com/facebook/CacheLib/assets/59281500/469a6375-28e3-46fc-a989-a53c6d43207a)

**Desktop (please complete the following information):**
 - Ubuntu 22.04.3 LTS
 - CacheLib tag: v20240312_RC

guaguastandup:
same question, same cachelib tag, I am using ubuntu22.04

> can you share more details regarding your linux setup? Which ubuntu version is this?

YuqingXiong:
> same question, same cachelib tag, I am using ubuntu22.04
> 
> > can you share more details regarding your linux setup? Which ubuntu version is this?

me too. I am using Ubuntu 22.04.3 LTS

YuqingXiong:
In fact, I have an inelegant solution.

I just commented the error output code, like this:

![image](https://github.com/facebook/CacheLib/assets/59281500/a5f64c66-0519-4454-b891-3733dc8426f0)

Then, return back Cachelib directory, and remove all build files: 
`rm -rf ./build-*` 

Finally, again to build, but not use git:
`./contrib/build.sh -d -j -v -S`

This error will not apper, but I think it might cause other bugs in the future.

MaoZiming:
Same problem

KIDSSCC:
I have frequently encountered issues with conflicts in the dynamically linked glog library. Based on my testing, a likely cause is that when building CacheLib, the -d option is used to compile third-party libraries in debug mode.

For example, in the case of glog, the CacheLib test programs attempt to link against the release version of glog during compilation. However, the CacheLib build list only includes the debug version. As a result, the compiler ultimately links against the system-wide release version of glog, while thriftcpp2 links against the debug version of glog. This mismatch leads to dependency conflicts.

I am unsure if this issue has been resolved in newer versions, but for v2023.04.24.00, it can be mitigated by running build.sh without the -d option.

pbhandar2:
Closing old issues. Please reopen or create new one if issue persists.

PR #314, opened 2024-05-22
Added Debug Flag To Examples Build.sh files to avoid conflict

aaditya2200:


therealgymmy:
Can you clarify what problem this resolves?

aaditya2200:
Without this, the examples do not build. I get an error regarding conflicting declarations. This is similar to this issue. https://github.com/facebook/CacheLib/issues/309

therealgymmy:
I cannot import this as it has conflicts with submodules. I'll send out a change from our internal repo that only contains the build.sh change

therealgymmy:
@aaditya2200: can you verify if the example now builds successfully?

Change landed at https://github.com/facebook/CacheLib/commit/d45bce26b86098c4915f4d13ab40d8304ccdfc5b

aaditya2200:
@therealgymmy Yes, this is the only change I have made. The builds work fine now.

PR #318, opened 2024-06-07
Adding Unit Tests for FdpNvme class

arungeorge83:
This commit adds the unit test cases for "FdpNvme" class.
It also adds a new constructor for FdpNvme. The existing constructor of FdpNvme communicates with the actual FDP device and makes the unit tests impossible to run without the actual device. So this commit adds a new constructor which allows to experiment with "FdpNvme" without the actual FDP device.

### DIFF
--- a/cachelib/navy/common/FdpNvme.cpp
+++ b/cachelib/navy/common/FdpNvme.cpp
@@ -31,8 +31,18 @@ namespace navy {
 
 FdpNvme::FdpNvme(const std::string& bdevName)
     : file_(openNvmeCharFile(bdevName)) {
-  initializeFDP(bdevName);
+  nvmeData_ = readNvmeInfo(bdevName);
+  Buffer buffer = nvmeFdpStatus();
+  struct nvme_fdp_ruh_status* ruh_status =
+      reinterpret_cast<struct nvme_fdp_ruh_status*>(buffer.data());
+  initializeFDPHandles(ruh_status);
   XLOGF(INFO, "Initialized NVMe FDP Device on file: {}", bdevName);
+}
+
+FdpNvme::FdpNvme(NvmeData& data, struct nvme_fdp_ruh_status* ruh_status)
+    : nvmeData_(data) {
+  initializeFDPHandles(ruh_status);
+  XLOGF(INFO, "Initialized NVMe FDP Device");
 }
 
 int FdpNvme::allocateFdpHandle() {
@@ -49,17 +59,11 @@ int FdpNvme::allocateFdpHandle() {
   return static_cast<int>(phndl);
 }
 
-void FdpNvme::initializeFDP(const std::string& bdevName) {
-  nvmeData_ = readNvmeInfo(bdevName);
-
-  Buffer buffer = nvmeFdpStatus();
-  struct nvme_fdp_ruh_status* ruh_status =
-      reinterpret_cast<struct nvme_fdp_ruh_status*>(buffer.data());
-
+void FdpNvme::initializeFDPHandles(struct nvme_fdp_ruh_status* ruh_status) {
   if (!ruh_status->nruhsd) {
     throw std::invalid_argument("Failed to initialize FDP; nruhsd is 0");
   }
-  placementIDs_.reserve(ruh_status->nruhsd);
+  placementIDs_.resize(ruh_status->nruhsd);
   maxPIDIdx_ = ruh_status->nruhsd - 1;
   for (uint16_t i = 0; i <= maxPIDIdx_; ++i) {
     placementIDs_[i] = ruh_status->ruhss[i].pid;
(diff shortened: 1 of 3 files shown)


haowu14 (approved):

cachelib/navy/common/tests/DeviceTest.cpp
```diff
@@ -196,6 +198,70 @@ TEST(Device, Stats) {
...
+// Test the Fdp Handle allocation
+TEST(FDP, FdpHandleAllocationTest) {
+  int nsId = 1;
+  uint32_t lbaShift = 12;
+  uint32_t maxTfrSize = 262144;
+  uint64_t startLba = 0;
+
+  NvmeData data(nsId, lbaShift, maxTfrSize, startLba);
+  struct nvme_fdp_ruh_status ruh_status;
+  ruh_status.nruhsd = 4;
```
haowu14, line 224: The ruh_status.ruhss is not initialized. This caused nullptr when running the test. can you please verify?

cachelib/navy/common/tests/DeviceTest.cpp
```diff
@@ -196,6 +198,70 @@ TEST(Device, Stats) {
...
+  folly::IoUringOp::Options options;
+  options.sqe128 = true;
+  options.cqe32 = true;
+
+  iouringCmdOp = std::make_unique<folly::IoUringOp>(
+      folly::AsyncBaseOp::NotificationCallback(), options);
+  iouringCmdOp->initBase();
+  struct io_uring_sqe& sqe = iouringCmdOp->getSqe();
+
+  void* buf;
```
haowu14, line 256: Please initialize this variable.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -31,10 +31,20 @@ namespace navy {
...
+  nvmeData_ = readNvmeInfo(bdevName);
+  Buffer buffer = nvmeFdpStatus();
+  struct nvme_fdp_ruh_status* ruh_status =
+      reinterpret_cast<struct nvme_fdp_ruh_status*>(buffer.data());
+  initializeFDPHandles(ruh_status);
   XLOGF(INFO, "Initialized NVMe FDP Device on file: {}", bdevName);
 }
 
+FdpNvme::FdpNvme(NvmeData& data, struct nvme_fdp_ruh_status* ruh_status)
+    : nvmeData_(data) {
```
haowu14, line 43: Please add a copy constructor for NvmeData class.

haowu14 (changes requested):
Comments based on running tests internally.

cachelib/navy/common/FdpNvme.cpp
```diff
@@ -49,13 +59,7 @@ int FdpNvme::allocateFdpHandle() {
...
-  nvmeData_ = readNvmeInfo(bdevName);
-
-  Buffer buffer = nvmeFdpStatus();
-  struct nvme_fdp_ruh_status* ruh_status =
-      reinterpret_cast<struct nvme_fdp_ruh_status*>(buffer.data());
-
+void FdpNvme::initializeFDPHandles(struct nvme_fdp_ruh_status* ruh_status) {
   if (!ruh_status->nruhsd) {
     throw std::invalid_argument("Failed to initialize FDP; nruhsd is 0");
   }
```
haowu14, line 65: Line 66 should be resize instead of reserve.

arungeorge83: Is this review comment related to only this PR or it is applicable to original FDP PR also? 
If it is applicable to FDP PR also, should we take this change seperately?

haowu14: I think this one needs to be changed otherwise the new test will be failing. When running the unit test internally I observed what seems like an index our of range fault and I suspected this was the problem.

cachelib/navy/common/tests/DeviceTest.cpp
```diff
@@ -196,6 +198,70 @@ TEST(Device, Stats) {
...
+  int nsId = 1;
+  uint32_t lbaShift = 12;
+  uint32_t maxTfrSize = 262144;
+  uint64_t startLba = 0;
+
+  NvmeData data(nsId, lbaShift, maxTfrSize, startLba);
+  struct nvme_fdp_ruh_status ruh_status{};
+  ruh_status.nruhsd = 4;
+
+  FdpNvme fdp = FdpNvme(data, &ruh_status);
```
haowu14, line 226: This is still crashing on me.
`ruh_status` has a field `ruhss`, which is an array of `nvme_fdp_ruh_status_desc` that's not initialized.

In `initializeFDPHandles`, we'll iterate through it as if it has `nruhsd` elements. Maybe initialize it here?

haowu14 (changes requested):

cachelib/navy/common/tests/DeviceTest.cpp
arungeorge83: @haowu14 We made some changes and made sure that there are no unitialized buffers w.r.t. ruh_status and ruhss. Hopefully this should solve the issue. Pls check it. 
(We could not reproduce the crash in our systems though)

PR #324, opened 2024-07-02
[not for commit] Change compressed pointer type to 64 bits

therealgymmy:
increase pointer size to 8 bytes so cachelib can address up to 1TB of memory


pbhandar2:
Thank you for the PR! This has been addressed with the introduction of 5 byte compressed pointer which should map more than 1 TB.

Issue #323, opened 2024-07-02
Build is failing with  error: ‘fmt::v10::detail::type_is_unformattable_for<const facebook::cachelib::navy::Status, char> _’ has incomplete type  1600 |     type_is_unformattable_for<T, typename Context::char_type> _;       |                                                               ^


sairahul-chalamalasetti:
**Describe the bug**
While building the cachelib with ./contrib/build.sh -j -T the build is failing with below error

error: ‘fmt::v10::detail::type_is_unformattable_for<const facebook::cachelib::navy::Status, char> _’ has incomplete type  1600 |     type_is_unformattable_for<T, typename Context::char_type> _;       |                                                               ^

**To Reproduce**
Clone the cachelib and run "./contrib/build.sh -j -T "

**Expected behavior**
The build the failing with already defined above error

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Ubuntu 22:04.4 LTS
Kernel: 6.9.0-060900-generic
g++ version: g++ (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0

fan982:
I have meet the same error, waiting for a solution.

tang-hi:
Put the code below in `cachelib/navy/common/Types.h`. It works for me.
```C++
inline auto format_as(Status status) {
  return fmt::underlying(status);
}
```

sairahul-chalamalasetti:
This fix helped to build on Kernel 6.9.

tang-hi:
I will try to raise a PR to merge it into the repo later

PR #328, opened 2024-07-21
use toString when print Status

tang-hi:
Fix this issue #323 
Should use `toString` when print Status


pbhandar2:
Thanks for the PR! Closing this since the recent build is good.

PR #351, opened 2024-11-19
fix compiler errors for main and tests

byrnedj:
This patch includes some of the changes in https://github.com/facebook/CacheLib/pull/349 to fix the compiling of the new .cpp files. 

It fixes some of the constexpr related errors.


pbhandar2:
Thanks for the PR! Closing this since most recent build is good.

PR #384, opened 2025-05-01
Fix CMake build related to index code refactor

c8ef:
In https://github.com/facebook/CacheLib/commit/d0139afcac157a0a3efa10dd833624da117ec5f0, `block_cache/Index.cpp` was renamed to `block_cache/SparseMapIndex.cpp`, but this change was not reflected in the cmake build, causing errors.


c8ef:
Hi @haowu14, could you please take a look at this patch? Thanks!

c8ef:
Hi @haowu14 @pbhandar2, could you please take a look at this patch? Thanks!

PR #418, opened 2025-11-20
Add S3FIFO Eviction Algorithm

williamnixon20:
This PR adds the S3FIFO eviction algorithm, as described [here](https://dl.acm.org/doi/pdf/10.1145/3600006.3613147).

Running cachebench results show that S3FIFO achieves the highest hit rate among all evaluated algorithms. Throughput wise, it performs similarly with TinyLFU.

Our implementation adapts the existing TinyLFUMM.

**Algorithm Detail**
The algorithm maintains three structures:
- S — a small queue for newly inserted items (10% of size)
- M — a main FIFO queue (90% of size)
- G — a ghost queue for items evicted from S (size of M), only tracks the object key hash.

The general idea of the algorithm is:
- Evict early: items in S are aggressively filtered to not pollute the main cache.
- Promote late: promotion/reinsertion occurs only when an accessed item reaches the tail.

Each object carries one access bit that records whether it has been accessed since it was inserted or last inspected.

Items in the tail are:
- If accessed and in S, moved to M.
- If accessed and in M, reinserted to head of M.
- If not accessed, evict. If this item is evicted from S, move it to the ghost queue.

Insertions are done as such:
- If the item exists in ghost, directly insert to M.
- Else insert to S.

We try to do eviction from S if the size > expected size (10%). Else, we try to evict from main.

Eviction iterator starts from tail and skips accessed items. Promotions (reinsertions) are done 1x before iterator is created.

**Benchmark:**
We evaluated S3FIFO using the test configurations in test_configs/hit_ratio with 48 threads and an 8 GB cache. We compared this against other existing MMContainers(LRU2Q, LRU, and TinyLFU).  We use the DebugRel release version for testing.

The results can be found here 
[S3 FIFO Benchmark (2).pdf](https://github.com/user-attachments/files/23684011/S3.FIFO.Benchmark.2.pdf)

**Miss ratio:**
S3FIFO consistently outperforms LRU, LRU2Q, and TinyLFU across all workloads.

**Throughput:**
Throughput is comparable to TinyLFU.

This implementation is not a throughput optimized version and still heavily uses LRU locks (in this case, TinyLFU's spinlocks). There maybe room for further improvement.

**Notes:**
- A FIFOHashSet is implemented to act as a ghost queue, using deque for maintaining FIFO order and unordered set for membership. 
- Several unused LRU configurations were removed as part of cleanup.

The parameters and the defaults are:
(1) Small / probationary queue size (default 10%)
Controls how long new objects stay in cache. If the workload has many low-reuse items (objects that appear only once or only a few times), making this smaller helps avoid pollution because these items get evicted quickly without entering the main cache.
(2) Large queue size (default 90%)
The rest of the cache. This is the “main” queue that holds objects with higher reuse.
(3) Ghost queue size (default 90%)
Stores only the key hashes of items evicted from the small queue. A larger ghost helps catch objects that return shortly after eviction. If the workload has many items that appear only twice and never come back again, the ghost queue can be made smaller since promoting these into the main queue doesn’t help.

For tuning, the paper shows that the small queue generally works best between 5% and 20% so those values can be tried out. If we have the proportion of the object reuse in the trace it can help us determine the sizes too.

image

### DIFF
--- a/cachelib/allocator/CacheAllocator.h
+++ b/cachelib/allocator/CacheAllocator.h
@@ -6077,6 +6077,7 @@ namespace facebook::cachelib {
 extern template class CacheAllocator<LruCacheTrait>;
 extern template class CacheAllocator<LruCacheWithSpinBucketsTrait>;
 extern template class CacheAllocator<Lru2QCacheTrait>;
+extern template class CacheAllocator<S3FIFOCacheTrait>;
 extern template class CacheAllocator<TinyLFUCacheTrait>;
 extern template class CacheAllocator<WTinyLFUCacheTrait>;
 
@@ -6097,6 +6098,14 @@ using LruAllocatorSpinBuckets = CacheAllocator<LruCacheWithSpinBucketsTrait>;
 //  3. items in cold queue are evicted to make room for new items
 using Lru2QAllocator = CacheAllocator<Lru2QCacheTrait>;
 using Lru5B2QAllocator = CacheAllocator<Lru5B2QCacheTrait>;
+
+// CacheAllocator with S3 FIFO eviction policy
+// It maintains 2 queues, one for for probation and one for the main queue.
+// New items are added to the probation queue, and if not accessed
+// will be quickly removed from the cache to remove 1 hit wonders.
+// If accessed while in probation, it will eventually be promoted to the main queue.
+// Items in the tail of main queue will be reinserted if accessed.
+using S3FIFOAllocator = CacheAllocator<S3FIFOCacheTrait>;
 
 // CacheAllocator with Tiny LFU eviction policy
 // It has a window initially to gauage the frequency of accesses of newly
--- a/cachelib/allocator/CacheAllocator.h
+++ b/cachelib/allocator/CacheAllocator.h
@@ -6106,6 +6106,7 @@ using Lru5B2QAllocator = CacheAllocator<Lru5B2QCacheTrait>;
 // If accessed while in probation, it will eventually be promoted to the main queue.
 // Items in the tail of main queue will be reinserted if accessed.
 using S3FIFOAllocator = CacheAllocator<S3FIFOCacheTrait>;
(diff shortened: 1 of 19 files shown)


williamnixon20:
> @williamnixon20 Could you also describe each configuration parameter of S3FIFO, the default values, when and how to tune the them in the summary?

The parameters and the defaults are:
(1) Small / probationary queue size (default 10%)
Controls how long new objects stay in cache. If the workload has many low-reuse items (objects that appear only once or only a few times), making this smaller helps avoid pollution because these items get evicted quickly without entering the main cache.
(2) Large queue size (default 90%)
The rest of the cache. This is the “main” queue that holds objects with higher reuse.
(3) Ghost queue size (default 90%)
Stores only the key hashes of items evicted from the small queue. A larger ghost helps catch objects that return shortly after eviction. If the workload has many items that appear only twice and never come back again, the ghost queue can be made smaller since promoting these into the main queue doesn’t help.


For tuning, the paper shows that the small queue generally works best between 5% and 20% so those values can be tried out. If we have the proportion of the object reuse in the trace it can help us determine the sizes too.

image

cachelib/allocator/CacheAllocator.h
```diff
@@ -6098,6 +6099,14 @@ using LruAllocatorSpinBuckets = CacheAllocator<LruCacheWithSpinBucketsTrait>;
 using Lru2QAllocator = CacheAllocator<Lru2QCacheTrait>;
 using Lru5B2QAllocator = CacheAllocator<Lru5B2QCacheTrait>;
 
+// CacheAllocator with S3 FIFO eviction policy
+// It maintains 2 queues, one for for probation and one for the main queue.
```
pbhandar2: Could you talk about the ghost queue as well?

pbhandar2:
submitted a new PR so closing this one

PR #461, opened 2026-05-26
[ci] Update GitHub Action mozilla-actions/sccache-action

gruenich:
Fixes deprecation warning for node20.js


rlyerly:
@gruenich thanks for the PR!  This action is generated from a template, so I'm going to appply your fix there & regenerate.

PR #485, opened 2026-09-05
cachelib: skip reaper traversal for empty caches

SimBuddy:
## Summary

The reaper can continue scanning assigned allocation slots after all accessible primary items have been removed. Use the existing primary-item count to skip that periodic slab walk when the cache is empty. Nonempty traversal and periodic wakeups are unchanged.

## Why

Assigned slabs can remain after the accessible population reaches zero, but walking those slots cannot find an accessible item to expire. The existing atomic count supplies the signal without another flag, lock, or per-item update.

## Concurrency

The count observation can race with insertion/publication. A concurrently inserted item may be handled on a subsequent periodic pass, consistent with the existing weak slab traversal. Foreground expiration checks and synchronized removal/ownership behavior are unchanged.

## Performance

In a local default-style idle workload (64 MiB configured cache, approximately 16 MiB assigned slots, 5-second reaper interval, 12-second window):

| Metric | Original | Patched |
|---|---:|---:|
| Median idle-process CPU | 5.44 ms | 0.27 ms |
| Median slot visits | 131,072 | 0 |

This saves about **5 ms CPU over the 12-second idle window**. It measures idle/background reaper work, not foreground cache throughput; the process CPU measurement includes other process bookkeeping. Absolute savings are modest and depend on assigned slab population, reaper frequency and hardware. A concise confirmation on the final rebased source verified zero empty-cache visits, active nonempty traversal and reactivation.

## Testing

- Current-upstream getdeps build and registered suite: **781 passed, 5 skipped, 1 existing disabled, 0 failures**.
- Three new registered tests cover allocated/unpublished/unlinked empty caches and chains; repeated reactivation, pinned expired lifetime and live retention; and deterministic concurrent slab traversal.
- Five unavailable huge-page tests were skipped; BlockCache.ExpiredItemDestructorCallback remains disabled upstream. Unregistered suites and TSAN are not claimed.
- `python3 build/fbcode_builder/getdeps.py --allow-system-packages --scratch-path /root/wob3/getdeps --num-jobs 4 build cachelib --src-dir /root/wob3/CacheLib --build-type RelWithDebInfo`
- `python3 build/fbcode_builder/getdeps.py --allow-system-packages --scratch-path /root/wob3/getdeps --num-jobs 2 test cachelib --src-dir /root/wob3/CacheLib --build-type RelWithDebInfo --no-testpilot --timeout 300 --retry 0` (private ext4 /tmp and DAC-bypass capabilities removed for filesystem/permission tests).
- `clang-format-21 --dry-run --Werror cachelib/allocator/CacheAllocator.h cachelib/allocator/tests/ReaperTest.cpp`
- `git diff --check upstream/main...HEAD`

## Scope

No cache or eviction policy changes, new persistent state, or NVM changes. No foreground hit/miss behavior change is intended. Numerical CPU results are workload and hardware dependent.

The opportunities were identified and validated using WOB/WOB3 techniques.

WOB is an experimental method of refocussing workloads in data-heavy code. I am stress testing various codebases, to develop the method further, and cachelib seemed like a worthwhile candidate. This single result was a small win, so I am sharing. Feedback would be greatly appreciated.

WOB is human-designed, human-led, and machine assisted.


cachelib/allocator/tests/ReaperTest.cpp
```diff
@@ -0,0 +1,166 @@
...
+  const auto deadline =
+      std::chrono::steady_clock::now() + std::chrono::seconds(5);
+  while (!predicate()) {
+    if (std::chrono::steady_clock::now() >= deadline) {
+      return false;
+    }
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  }
+  return true;
+}
```
rlyerly: Please use [eventuallyTrue](https://github.com/facebook/CacheLib/blob/main/cachelib/common/TestUtils.h#L30) or [ASSERT_EVENTUALLY_TRUE](https://github.com/facebook/CacheLib/blob/main/cachelib/common/TestUtils.h#L49) instead.

cachelib/allocator/tests/ReaperTest.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+TEST(ReaperTest, ReactivatesAcrossEmptyTransitions) {
+  Cache cache(reaperTestConfig());
+  const auto pool =
+      cache.addPool("pool", cache.getCacheMemoryStats().ramCacheSize);
+  Reaper<Cache> reaper(cache, {});
+  ASSERT_TRUE(reaper.start(std::chrono::milliseconds(10), "reaper-test"));
+  ASSERT_TRUE(waitFor([&] { return reaper.getStats().numTraversals > 0; }));
+  EXPECT_EQ(reaper.getStats().numVisitedItems, 0);
+  for (size_t i = 0; i < 20; ++i) {
```
rlyerly, line 96: What's the point of doing this 20 times?  Seems like 2 times is enough to guarantee we go between skipping & traversing

cachelib/allocator/tests/ReaperTest.cpp
rlyerly, line 49: Please use [eventuallyTrue](https://github.com/facebook/CacheLib/blob/main/cachelib/common/TestUtils.h#L30) or [ASSERT_EVENTUALLY_TRUE](https://github.com/facebook/CacheLib/blob/main/cachelib/common/TestUtils.h#L49-L50) instead.

cachelib/allocator/tests/ReaperTest.cpp
```diff
@@ -0,0 +1,166 @@
...
+    EXPECT_EQ(cache.getAccessContainerNumKeys(), 1);
+    cache.remove("live");
+  }
+  reaper.stop(std::chrono::seconds(0));
+  EXPECT_EQ(reaper.getStats().numReapedItems, 20);
+  EXPECT_EQ(reaper.getStats().numVisitErrs, 0);
+  EXPECT_EQ(cache.getAccessContainerNumKeys(), 0);
+}
+
+TEST(ReaperTest, NonemptyCacheStillSkipsSlabDuringConcurrentTraversal) {
```
rlyerly, line 130: Not exactly sure what this test is trying to accomplish, can you give me some details?

cachelib/allocator/CacheAllocator.h
```diff
@@ -2194,6 +2194,13 @@ class CacheAllocator : public CacheBase {
   // time.
   template <typename Fn>
   void traverseAndExpireItems(Fn&& f) {
+    // Skip the slab walk when there are no accessible primary items. An
+    // insertion racing this snapshot can be handled by the next periodic pass,
+    // as with the existing weakly consistent slab walk.
```
rlyerly, line 2199: Delete the the second sentence, too much irrelevant detail.  Also by "primary" do you mean either individual items or parents of chained items?  Not aware of that terminology, I think we can delete it.
```
// Skip the slab walk when there are no accessible items
```

rlyerly (changes requested):
Few smaller cleanups & questions.  Thanks!

cachelib/allocator/tests/ReaperTest.cpp
SimBuddy: Much of the finer detail of this is lost on me - the system I am developing is guided by me, but the output is often machine-only - my intention is to test on many different codebases as I develop it - I've had good feedback thus far, but I'm unable to elaborate much - if the amendments are of no use, that's feedback too, and I will take on board.  This particular codebase produced almost nothing in the way of potential enhancements beyond this suggestion.

rlyerly: To be frank, I expect this to have minimal impact on most deployments.  This elides work when caches are empty, but almost no real production user of Cachelib has an empty cache.  Closing for now, feel free to reopen if you want to drive this further.

PR #489, opened 2026-09-15
cmake: prefer glog's own CMake config over the FindGlog module

michel-slm:
glog >= 0.7 requires consumers to define GLOG_USE_GLOG_EXPORT, which its exported glog::glog target carries as an INTERFACE_COMPILE_DEFINITION. Our FindGlog module constructs its own imported glog::glog target from find_library/find_path results and never sets that definition, so on a distro shipping glog 0.7 (Fedora 44 has 0.7.1) every cachelib translation unit fails with:

  /usr/include/glog/logging.h:1228:30: error: expected primary-expression before 'public'
  folly/io/Cursor.h:834:5: error: 'LogMessageFatal' is not a member of 'google'

Try find_package(glog CONFIG) first and only fall back to the manual search when no config file is installed. getdeps' from-source glog also ships a config file, so the fallback now only matters for glog builds installed without CMake support.

Tested on Fedora 44 in a single `getdeps.py --allow-system-packages build --no-tests cachelib` run together with the Fedora rpm mappings from facebook/CacheLib#488 and the manifest change that honours --no-tests for cachelib; the build succeeds against glog-devel 0.7.1 with glog_DIR=/usr/lib64/cmake/glog in the CMake cache.


michel-slm:
## Build evidence for review

Environment: Fedora 44 toolbox (aarch64), `glog-devel-0.7.1-1.fc44`, `cmake 4.3.0`, getdeps with `--allow-system-packages`. The cachelib tree also carried #488 (Fedora rpm mappings) and #490 (`--no-tests` honoured for cachelib) during these runs.

### (a) System glog ≥ 0.7 available: config path is taken and the build succeeds

Command (from the CacheLib checkout):

```
python3 build/fbcode_builder/getdeps.py --allow-system-packages --num-jobs 3 \
  --extra-cmake-defines '{"CMAKE_POLICY_VERSION_MINIMUM":"3.5"}' \
  build --free-up-disk --src-dir=. --no-tests cachelib
```

CMake cache after configure:

```
glog_DIR:PATH=/usr/lib64/cmake/glog
```

`CMakeConfigureLog.yaml` records the new call site resolving glog in config mode:

```
kind: "find_package-v1"
  - "cmake/FindGlog.cmake:29 (find_package)"
  - "CMakeLists.txt:112 (find_package)"
name: "glog"
configs:
  - filename: "glog-config.cmake"   kind: "cmake"
  ...
  mode: "config"
```

`build.ninja` carries the definition that glog 0.7 requires, inherited from the exported target:

```
$ grep -c GLOG_USE_GLOG_EXPORT build/cachelib/build.ninja
200
```

Build result: exit 0, `installed/cachelib/.built-by-getdeps` written, final steps

```
[208/212] Linking CXX executable cachebench/cachebench
[211/212] Linking CXX executable cachebench/binary_trace_gen
```

and `ldd cachebench` shows `libglog.so.2 => /lib64/libglog.so.2`.

For comparison, the same build on the same host **without** this change (FindGlog module constructing its own `glog::glog` target) failed on every translation unit:

```
/usr/include/glog/flags.h:105:1: error: 'GLOG_EXPORT' does not name a type
/usr/include/glog/logging.h:1229:2: error: expected primary-expression before 'public'
folly/io/Cursor.h:834:5: error: 'LogMessageFatal' is not a member of 'google'; did you mean 'LogMessageTime'?
```

### (b) Fallback path (no glog CMake config) still works

Same environment, fresh build directory, config mode forcibly disabled so the module's `find_path`/`find_library` branch runs:

```
cmake <same defines as getdeps> -DCMAKE_DISABLE_FIND_PACKAGE_glog=TRUE
```

Configure exits 0 and takes the fallback:

```
-- Found glog: /usr/lib64/libglog.so
```

```
CMAKE_DISABLE_FIND_PACKAGE_glog:UNINITIALIZED=TRUE
GLOG_INCLUDE_DIR:PATH=/usr/include
GLOG_LIBRARY:FILEPATH=/usr/lib64/libglog.so
$ grep -c GLOG_USE_GLOG_EXPORT build.ninja
0
```

So discovery through the module is unchanged. Compiling one object in that configuration against glog 0.7.1 fails, as expected and exactly as before this change, because the module path cannot know about the export define:

```
/usr/include/glog/logging.h:60:4: error: #error <glog/logging.h> was not included correctly. See the documentation for how to consume the library.
```

That is the pre-existing behaviour this PR fixes for the config case; the fallback remains sufficient for glog < 0.7 and for glog installs that ship no CMake config. getdeps' own from-source glog (pinned pre-0.7) installs a config file, so it also goes through the new config path.

Pre-existing and unchanged: CMake's warning that `find_package_handle_standard_args(glog ...)` does not match the module name `Glog` (`cmake/FindGlog.cmake:38`).

Full logs (11 MB online build log, configure logs, CMake caches) are preserved locally and can be shared on request.

michel-slm:
## Follow-up: documented output variables on the config path

Review nit addressed: the early return after `find_package(glog CONFIG)` used to skip setting `GLOG_INCLUDE_DIR`, `GLOG_INCLUDE_DIRS` and `GLOG_LIBRARIES`. They are now derived from the imported target, with `GLOG_LIBRARIES` set to `glog::glog` so that variable-style consumers still inherit `GLOG_USE_GLOG_EXPORT`. The config check also moved above the unconditional `find_library`, and the header lists the target as an output.

Probe: a minimal project that only uses the variables (`target_link_libraries(probe ${GLOG_LIBRARIES})`, `target_include_directories(probe PRIVATE ${GLOG_INCLUDE_DIRS})`) with a `LOG(INFO)` call, Fedora 44, glog-devel 0.7.1.

Config path:

```
-- GLOG_FOUND=TRUE
-- GLOG_INCLUDE_DIR=/usr/include
-- GLOG_INCLUDE_DIRS=/usr/include
-- GLOG_LIBRARIES=glog::glog
[100%] Linking CXX executable probe
[100%] Built target probe
```

Fallback path (`-DCMAKE_DISABLE_FIND_PACKAGE_glog=TRUE`), unchanged behaviour:

```
-- Found glog: /usr/lib64/libglog.so
-- GLOG_FOUND=TRUE
-- GLOG_INCLUDE_DIR=/usr/include
-- GLOG_INCLUDE_DIRS=/usr/include
-- GLOG_LIBRARIES=/usr/lib64/libglog.so
```

PR #490, opened 2026-09-15
getdeps: honour --no-tests for cachelib

michel-slm:
cachelib's CMakeLists defaults BUILD_TESTS to ON and the manifest never overrode it, so `getdeps.py build --no-tests cachelib` still compiled every test binary. Wire the getdeps test context to BUILD_TESTS the same way the folly and fizz manifests do.

With tests off the cachelib build on Fedora 44 drops from 497 to 110 ninja steps, which matters for memory- and CPU-constrained distro builders.

Tested on Fedora 44 in a single `getdeps.py --allow-system-packages build --no-tests cachelib` run together with the Fedora rpm mappings from facebook/CacheLib#488 and the FindGlog config-mode fix; the resulting CMake cache has BUILD_TESTS:BOOL=OFF and the build succeeds.


michel-slm:
## Verification of both directions of the `test` context

Environment: Fedora 44, getdeps with `--allow-system-packages`, on a tree that also carried #488, #489 and the vendoring branch; deps already built, so only cachelib's own configure step ran each time.

### `--no-tests` passed → `BUILD_TESTS=OFF`

From the full offline build earlier today (`build --free-up-disk --no-tests --src-dir=. cachelib`, exit 0):

```
$ grep BUILD_TESTS build/cachelib/CMakeCache.txt
BUILD_TESTS:BOOL=OFF
```

cachelib compiled in 221 ninja steps and no `*-test-*` executables were produced.

### `--no-tests` omitted → `BUILD_TESTS=ON`

Same scratch dir, same options minus `--no-tests`, with a no-op CMake target so that only the configure step runs:

```
python3 build/fbcode_builder/getdeps.py --allow-system-packages ... \
  build --no-deps --cmake-target help --src-dir=. cachelib
```

getdeps passed the define from the new `[cmake.defines.test=on]` section on the cmake command line:

```
$ grep -oE 'DBUILD_TESTS=[A-Z]+' build.log | sort | uniq -c
      1 DBUILD_TESTS=ON
```

and the cache flipped accordingly:

```
$ grep BUILD_TESTS build/cachelib/CMakeCache.txt
BUILD_TESTS:BOOL=ON
```

After that configure, ninja knows about 212 `*-test-*` executable targets, and a dry run lists 276 remaining steps (the test binaries) on top of the already-built non-test objects, matching the 497-step total seen before this change.

### CI implication

The GitHub Actions build step invokes `getdeps.py build ... cachelib` without `--no-tests`, so it lands in the `test=on` branch and keeps building the tests that the subsequent `getdeps.py test` step runs (783 passing on today's main). Only the `--no-tests` path changes behaviour, from silently building everything to actually skipping the tests.

PR #493, opened 2026-09-16
getdeps: bump sparsemap to 0.7.0 for CMake 4

michel-slm:
sparse-map 0.6.2 declares cmake_minimum_required(VERSION 3.1). CMake 4 removed compatibility with anything below 3.5, so on hosts whose system CMake is 4.x (Fedora 44, Homebrew, the GitHub macOS runners, and soon Ubuntu 26.04 and Debian 14) the sparsemap step fails to configure:

```
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.
```

Upstream fixed this in 0.7.0 (released 2025-11-02), which requires CMake 3.10, so take the release rather than carry a patch. 0.7.0 also moves to C++17, fixes iterator invalidation on a non-inserting insert and the serialization of small hashes, and adds a no-exceptions mode; the headers install to the same include/tsl layout. cachelib is the only manifest that depends on sparsemap, and the one internal type it reaches into, tsl::detail_sparse_hash::sparse_array, keeps the same template parameters.

Tested on Fedora 44 with system CMake 4.3.0 and gcc 15, no CMAKE_POLICY_VERSION_MINIMUM anywhere on the command line: `getdeps.py --allow-system-packages build --no-tests --src-dir=. cachelib` configured and installed sparsemap 0.7.0 with no CMake error (the only CMake output of note being the < 3.10 deprecation warnings from fizz and wangle, tracked in facebook/folly#2703), compiled SparseMapIndex.cpp against the new headers with -Werror on, and linked cachebench; 51 minutes at two jobs.

Manifest-only change; follow-up to #488/#490/#491. The CMake version matrix behind the platform list is in facebook/folly#2703.


bigfootjon (approved):
Review automatically exported from Phabricator review in Meta.

PR #498, opened 2026-09-18
cachebench: link binary_trace_gen against cachelib_cachebench

michel-slm:
## Summary

`cachelib_binary_trace_gen` is a second library built from a subset of the `cachelib_cachebench` sources. It lists `runner/Runner.cpp` and `runner/Stressor.cpp` but not the `runner/`, `consistency/` and `workload/` files those two call into. With static libraries the linker only pulls the objects `binary_trace_gen.cpp` uses, so the gap never showed. With `BUILD_SHARED_LIBS=ON` (getdeps `--shared-lib`) the whole library becomes a `.so` with 36 undefined references and the executable fails to link. Even if it linked, the install rules never install that library, so the installed `binary_trace_gen` would not run. The duplicate also misses the `SKIP_OPTION_SIZE_VERIFY` definition applied to `cachelib_cachebench` on non-x86_64.

This drops the duplicate and links `binary_trace_gen` against `cachelib_cachebench`, which already contains every source it listed. 1 insertion, 25 deletions.

## Failure before the fix

Fedora Rawhide aarch64 package build (getdeps `--shared-lib`, GCC 15, `-flto=auto`), 36 undefined references, e.g.:

```
[227/228] Linking CXX executable cachebench/binary_trace_gen
FAILED: cachebench/binary_trace_gen
/usr/bin/ld.bfd: cachebench/libcachelib_binary_trace_gen.so: undefined reference to `typeinfo for facebook::cachelib::cachebench::CacheStressorBase'
/usr/bin/ld.bfd: cachebench/libcachelib_binary_trace_gen.so: undefined reference to `facebook::cachelib::cachebench::CacheStressorBase::finish()'
/usr/bin/ld.bfd: cachebench/libcachelib_binary_trace_gen.so: undefined reference to `facebook::cachelib::cachebench::ValueTracker::ValueTracker(std::vector<folly::Range<char const*>, ...>)'
/usr/bin/ld.bfd: cachebench/libcachelib_binary_trace_gen.so: undefined reference to `facebook::cachelib::cachebench::ProgressTracker::~ProgressTracker()'
/usr/bin/ld.bfd: cachebench/libcachelib_binary_trace_gen.so: undefined reference to `facebook::cachelib::cachebench::SimpleFlashBenchmarkGenerator::SimpleFlashBenchmarkGenerator(facebook::cachelib::cachebench::StressorConfig const&)'
collect2: error: ld returned 1 exit status
ninja: build stopped: cannot make progress due to previous errors.
```

## Test plan

Applied the patched `cachelib/cachebench/CMakeLists.txt` to the same build tree and re-ran only the failing target; CMake reconfigured and the link succeeded:

```
$ ninja -j 4 binary_trace_gen
[1/2] Re-running CMake...
[2/2] Linking CXX executable cachebench/binary_trace_gen
$ ls -la cachebench/binary_trace_gen
-rwxr-xr-x 1 root root 31564080 Sep 18 11:14 cachebench/binary_trace_gen
$ ldd cachebench/binary_trace_gen | grep cachelib
        libcachelib_cachebench.so.19 => not found   # build tree, not yet installed
        libcachelib_interface.so.19 => not found
        libcachelib_common.so.19 => not found
```

The `not found` entries are the uninstalled build tree; `cachebench` itself reports the same three. A full Fedora Rawhide package build with this patch applied is in progress and I will confirm here once it has completed `%install`.

Static builds are unaffected: `cachelib_cachebench` already contained every source the dropped library listed, and `binary_trace_gen.cpp` only needs `KVReplayGenerator.h` (header-only), `Config` and `Utils`.


michel-slm:
Runtime verification from the Fedora package built with this change (Rawhide aarch64, statically linked executables, installed from the RPM into a clean chroot):

```
$ cachelib_binary_trace_gen --help          # gflags exits 1 after printing help
    -json_test_config (path to test config. If empty, use default setting)
      type: string default: ""
    -progress (if set, prints progress every X seconds as configured, 0 to
      disable) type: uint64 default: 60

$ cp -r /usr/share/cachelib/test_configs/trace_replay/kvcache /tmp/kv && cd /tmp/kv
$ cachelib_binary_trace_gen --json_test_config config_kvtrace_gen_binary.json
Binary Trace Generator
I0918 17:58:46 ReplayGeneratorBase.h:124] Opened trace file /tmp/kv/kv_traces_1.csv
I0918 17:58:46 KVReplayGenerator.h:131] Started generating binary file from KVReplayGenerator(amp factor 1, size factor 1, trace requests 99)
$ ls -la kv_trace.bin
-rw------- 1 root root 2288 Sep 18 17:58 kv_trace.bin

$ cachebench --json_test_config config_kvtrace_replay_binary.json --progress 5
...
Total Ops : 2.00 million
I0918 17:58:52 Cache.h:1060] Shut down succeeded. Metadata is at: /tmp/cachelib_metadata
Finished!
```

The binary replay of this 99-request trace finishes with exit 0 but its stats report zero processed samples and no allocations; the CSV replay of the same trace (`config_kvtrace.json`) shows a live cache (24 allocations, 23 items in RAM, 2.74% hit ratio), so the executables themselves are fine. Whether the binary replay path is expected to behave that way on a trace this small is a separate question I have not looked into.

(The full-build confirmation took a detour: the first shared-library build linked but every `libcachelib_*.so` was unloadable, see #499, #500 and facebook/fbthrift#712. The Fedora package now links the executables statically.)

PR #499, opened 2026-09-18
getdeps: keep $LDFLAGS on the shared library links under --shared-lib

michel-slm:
## Summary

`--shared-lib` (#491) appends `-Wl,--exclude-libs=ALL` by defining `CMAKE_SHARED_LINKER_FLAGS` and `CMAKE_MODULE_LINKER_FLAGS` on the cmake command line. A cache variable given on the command line is never seeded from `$LDFLAGS`, so the caller's linker flags silently vanished from every shared library link, while the executables (whose `CMAKE_EXE_LINKER_FLAGS` we leave alone) kept them.

Seed both variables from `$LDFLAGS`, as CMake itself would, before appending. An explicit caller-supplied value still wins.

## How it showed up

Fedora package build of cachelib with `--shared-lib`, `LDFLAGS` set by rpm's `%set_build_flags`. From the generated `build.ninja`:

```
# executable: cachebench
LINK_FLAGS = -Wl,-z,relro -Wl,--as-needed -Wl,-z,pack-relative-relocs -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld ...

# shared library: libcachelib_common.so.19.1.0
LINK_FLAGS = -shared -Wl,--exclude-libs=ALL -Wl,--dependency-file=...
```

The libraries lost RELRO, BIND_NOW and, on Fedora Rawhide, `--error-rwx-segments`. That last one would have turned a real problem into a link error: fbthrift's generated `_data.cpp` files put relocated data in a `.rodata.*` section, which `-fPIC` makes writable, so every `libcachelib_*.so` was linked with a single RWX LOAD segment (ld only warned) and glibc's aarch64 loader then crashed on them at startup. With this fix the hardening flags reach the `.so` links and that class of problem fails loudly at link time.

## Test plan

```
$ cd build/fbcode_builder && python3 -m unittest getdeps.test.shared_lib_test
Ran 16 tests in 0.006s
OK
```

Two new tests: `$LDFLAGS` seeds both variables and `--exclude-libs=ALL` is appended; an explicit `CMAKE_SHARED_LINKER_FLAGS` from the caller overrides `$LDFLAGS` for that variable only. `black --check` clean.

Fedora Rawhide aarch64 package rebuild with this patch applied is in progress; I will confirm the shared libraries' `LINK_FLAGS` here once it completes.


michel-slm:
Verified with a full Fedora Rawhide aarch64 rebuild of CacheLib using `--shared-lib` with this patch applied. From the generated `build.ninja`, the shared libraries now get the same flags as the executables:

```
# libcachelib_common.so.19.1.0, before
LINK_FLAGS = -shared -Wl,--exclude-libs=ALL -Wl,--dependency-file=...
# after
LINK_FLAGS = -shared -Wl,-z,relro -Wl,--as-needed -Wl,-z,pack-relative-relocs -Wl,-z,now -specs=/usr/lib/rpm/redhat/redhat-hardened-ld -specs=/usr/lib/rpm/redhat/redhat-hardened-ld-errors ... -Wl,--exclude-libs=ALL
```

Every `libcachelib_*.so` came out with separate R E and RW `LOAD` segments plus `GNU_RELRO`, and `--error-rwx-segments` was in effect for the links (zero RWX warnings once facebook/fbthrift#712 was applied). Marking ready for review.

